PR #8, opened 2018-06-27
docker-compose changes

romainchassaigne:
I have changed something on the docker-compose file so that way you don't have to connect through the host network but only thanks to the stack network. In configuration, just put `mysql`for DB instead of `127.0.0.1` and `php:9000`


laszlo462:
I like this, exactly what I was looking for early on but never got around to changing.  Keeps the stack nice and tidy.

cdujeu:
This PR refering to php is probably deprecated - reopen if needed.

PR #54, opened 2019-01-25
Fix private IP bug

mzch:
Fix a bug that needs nowhere private ip

### DIFF
--- a/vendor/github.com/micro/go-web/extractor.go
+++ b/vendor/github.com/micro/go-web/extractor.go
@@ -53,7 +53,8 @@ func extractAddress(addr string) (string, error) {
 	}
 
 	if ipAddr == nil {
-		return "", fmt.Errorf("No private IP address found, and explicit IP not provided")
+		//return "", fmt.Errorf("No private IP address found, and explicit IP not provided")
+		return "127.0.0.1", nil
 	}
 
 	return net.IP(ipAddr).String(), nil


cdujeu:
Hi @mzch 
Thanks for contributing! I'm not sure we want to modify the content of the vendor/ folder, as it should kept synced at least with a certain branch of the original repository.
@ghecquet is there anyway to disable this check at the Cells level?

cdujeu:
Hi @mzch we should indeed switch to go modules at one point.
Quick question, would you be interested in participating to translation of Pydio Cells into Japanese?

cdujeu:
We're on Crowd In https://crowdin.com/project/pydio-cells
I think we would have to create a new language, can you sign up (free) and see if you can "propose" a new language?

mzch:
I logged in Crowd and requested to add Japanese language.

bsinou:
Japanese has been added for Pydio Cells in crowdin. Thanks for helping and don't hesitate if you have any question.

bsinou:
Hey @mzch, just for the record, I had to remove and re-add the Japanese language in the crowdin project to insure generated file names respect our internal conventions. But now everything is up and running. So feel free to jump in at any time.

cdujeu:
Finally fixed in 1.4.1 => pass an address via services options instead of hacking the lib.

PR #59, opened 2019-02-08
Fix: IDM groups can not be deleted

fajo-de:
(see also: https://forum.pydio.com/t/unable-to-delete-user-groups-1-2-5-dev-head/2111)

This PR fixes (or works around?) an issue where IDM groups can not be deleted. In code a group deletion request gets differentiated from a user request by a request path suffix. In code the suffix "%2F" was required that for but during tests the browser used to append "/" (unencoded) instead causing group deletion requests to fail.

The only change made in this PR is to also allow "/" as request suffix ("/" is not allowed in group IDs).

### DIFF
--- a/cmd/ctl/cmd/user.go
+++ b/cmd/ctl/cmd/user.go
@@ -94,7 +94,7 @@ func searchUser(ctx context.Context, cli idm.UserServiceClient, login string) ([
 func deleteUser(ctx context.Context, login string) error {
 
 	singleQ := &idm.UserSingleQuery{}
-	if strings.HasSuffix(login, "%2F") {
+	if strings.HasSuffix(login, "%2F") || strings.HasSuffix(login, "/") {
 		// log.Logger(ctx).Debug("Received User.Delete API request (GROUP)", zap.String("login", login))
 		singleQ.GroupPath = login
 		singleQ.Recursive = true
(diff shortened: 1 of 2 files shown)


cdujeu:
mmm, that's interesting, what's your browser? do you have any proxy between cells and the browser?

fajo-de:
I tried different combinations (Linux: Chrome, Firefox; Windows: Edge) - all with the same result.

Yes, a proxy (apache based) is running in between. The logs show:
`[08/Feb/2019:13:02:24 +0100] "DELETE /a/user/test-group%2F HTTP/1.1" 200 58`

Which would be correct ...
 
The proxy should, however, not manipulate the request URI but pass it as is. "AllowEncodedSlashes On" is set for the VHost but should not have any effect when proxying - the behaviour will not change when switching it off.

Just being curious, is there a particular reason for using %2F instead of "/"? From my understanding it should not interfere with any functionality.

fajo-de:
... update: AllowEncodedSlashes Off (the default) would cause the proxy to respond with 404 (which he did). So it's required to be switched on to allow encoded slashes.

cdujeu:
So basically if you set `AllowEncodedSlashes On` the %2F value is working without your fix? 
The reason of this %2F is the js api client that we use, it is generated and uri-encodes parameters at one point (we can't prevent this).

I think we can merge the fix anyway, it does not cost much ;-)

Charles

fajo-de:
With "AllowEncodedSlashes On" it does not work - the request is passed on to cells but causes it to be rewritten to "/" somehow/somewhere. With it set to "Off" the proxy will block the request and will not pass it on to cells. So, good to have the fix merged - this will prevent others (and me) from having this issue in a proxy configuration.

PR #60, opened 2019-02-09
Eleminate use of depreciated tx_isolation for MySQL 5.7.20+ compatibility

fajo-de:
Eleminate use of depreciated tx_isolation in favour of 'SET SESSION TRANSACTION ISOLATION LEVEL' for MySQL 5.7.20+ compatibility.

The system variable _tx_isolation_ has been marked depreciated in MySQL 5.7.20 and a new alias _transaction_isolation_ has been added. This will break cell for MySQL 5.7.20+.

To gain compatibility to pre 5.7.20 and 5.7.20+ the DSN parameter tx_isolation has been replaced by executing the SQL statement "SET SESSION TRANSACTION ISOLATION LEVEL" prior to calling migrate().

### DIFF
--- a/vendor/github.com/coreos/dex/storage/sql/config.go
+++ b/vendor/github.com/coreos/dex/storage/sql/config.go
@@ -175,7 +175,6 @@ func (s *MySQL) open(logger logrus.FieldLogger) (*conn, error) {
 
 	params := make(map[string]string)
 	params["autocommit"] = "false"
-	params["tx_isolation"] = "SERIALIZABLE"
 	dexDSN.Params = params
 
 	db, err := sql.Open("mysql", dexDSN.FormatDSN())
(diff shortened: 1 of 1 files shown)


cdujeu:
Hello @fajo-de 
Basically we **never** commit directly inside the `vendor/` folder: this folder provides a way to keep dependencies consistent across builds, and it is only updated when the original dependency is updated. 
As such, we should update the original coreos/dex répo (wich is in fact forked in `pydio/dex`) and then update vendor with govendor tool. 
We'll have a look at that.
Best
Charles

fajo-de:
Hi Charles,
I checked the official coreos/dex repo but it appears to not even support MySQL (I only found PostgreSQL and SQLite3). So my thought was, that you use a, more or less, standalone fork of it. Shall I close the PR then (I can not create a PR on the pydio/dex fork without forking it) ?

fajo-de:
Appears to have been fixed in commit 76b356415481f8b9c817de8f80fe70f2a3b252bc and therefore can be closed.

cdujeu:
Fixed in 1.4.1

PR #106, opened 2019-04-12
Translate new messages in French and German

bsinou:
Thanks @fajo-de for the help with German messages.
Yet, I had to adapt some of them, so a double check is welcomed!


fajo-de:
Hi @bsinou I updated some of the German translations. You may want to update the language file.

PR #200, opened 2020-02-18
doc: External port syntax

drzraf:


cdujeu:
Hi, thx - but to be really accurate, shouldn't we use HOST instead of PORT in that case?

bsinou:
Or even rather FQDN see [1] or do you thinkj it os too cryptic for non sysadmin user that will install Cells ? 

[1]: https://en.wikipedia.org/wiki/Fully_qualified_domain_name

drzraf:
Honestly, I'm not [completely sure about the whole topic](https://forum.pydio.com/t/url-doesn-t-open-during-cells-installation-process/2954/12)
In particular, URL does not express whether an IP address is valid. But since some parameters may be extracted for use with LetsEncrypt, it's an important distinction.

PR #248, opened 2020-06-22
Reconciliate i18n branch for easy git workflows

bsinou:


bsinou:
Legacy PR. Not to be used. Closing without merging.

PR #268, opened 2020-10-24
Add ability to specify S3 endpoint during installation

lsde:
Replace hardcoded "s3.amazonaws.com" S3 endpoint during installation via CLI with a prompt, thus enabling Pydio to use other S3 servers for primary storage.


lsde:
I haven't included Grunt-generated static files in the commit, as I don't know your policy about those files, but I'm happy to commit those also.

cdujeu:
Hello @lsde, thanks for contributing! 
As a matter of fact, we have important changes coming soon for supporting multiple bind addresses in a beautiful way. A lot of work has been done on the `multi-bind` branch, and specifically the install parts have seen a lot of changes. The main idea is to avoid asking for bind/external at startup (binding to 0.0.0.0 in self-signed mode by default), and then to provide the ability to add one or more "site" with the `cells config sites` new command. 
If you have some time, could kindly try to checkout this branch, test it, test the status of full-s3 setup, and re-adapt your changes on this branch? If not no worries, i will do the merge later on (but basically the master is currently almost "frozen"). 
As for JS build files, we do commit them. It's not a perfect approach but is really more productive for continuous integration (as building each and every plugin of the UX from scratch takes a lot of time). 
Best
Charles

lsde:
Nice! I've commited also the static files, and merging my changes into `multi-bind`branch went without issues, and installation of this version finishes sucessfully both via CLI and GUI.
So, would you like me to switch the base to `multi-bind`?
I personally think commiting those into master and then merging it to this branch might be a better solution, but leave that up to you as a developer 😄

cdujeu:
hi @lsde sorry for coming back later on this, did not take time to handle it. 
`multi-bind` is now officially merged into master :-)
Can you check a last time this PR is correct and can be merged into master? also please make sure to sign the cla if not already done (see website => contribute pages)

cdujeu:
Manually reported in master, plus including custom S3 region field as well. Avail in next release :-)

PR #322, opened 2021-05-10
Update vendor/github.com/jaytaylor/go-hostsfile

skozlov404:
New version fixes build instructions that allows build on FreeBSD.


cdujeu:
hello,
Thanks for posting the request.
We do not update vendor folder directly (but by using govendor tool, which will update vendor.json as well). 
Can you instead open an issue asking for the update ? 
Thx

skozlov404:
Filed issue instead: #324

PR #326, opened 2021-05-26
Use Linux-like default workdir on FreeBSD

skozlov404:
FreeBSD, like Linux, prefers non-capitalized names, so use "pydio"
instead of "Pydio" for the vendor when choosing the default workding
directory.


cdujeu:
Applying manually in new `main` branch.

PR #488, opened 2023-05-27
Remove `github.com/ghodss/yaml` dependency

Juneezee:
At the time of making this commit, the package `github.com/ghodss/yaml` is no longer actively maintained.

We are only using `github.com/ghodss/yaml` in `common/auth/mapping-rule.go`. We can use `gopkg.in/yaml.v2` instead.

### DIFF
--- a/common/auth/mapping-rule.go
+++ b/common/auth/mapping-rule.go
@@ -26,16 +26,16 @@ import (
 )
 
 type MappingRule struct {
-	RuleName string
+	RuleName string `yaml:"RuleName"`
 
 	// Left Attribute is attribute of external user (ldap, sql, api ...)
 	// For example: displayName, mail, memberOf
-	LeftAttribute string
+	LeftAttribute string `yaml:"LeftAttribute"`
 
 	// Right Attribute is attribute of standard user
 	// For example: displayName, email
 	// Two reserved attributes: Roles, GroupPath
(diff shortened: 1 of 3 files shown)


common/auth/mapping-rule.go
```diff
@@ -26,16 +26,16 @@ import (
...
 	// Left Attribute is attribute of external user (ldap, sql, api ...)
 	// For example: displayName, mail, memberOf
-	LeftAttribute string
+	LeftAttribute string `yaml:"LeftAttribute"`
 
 	// Right Attribute is attribute of standard user
 	// For example: displayName, email
 	// Two reserved attributes: Roles, GroupPath
-	RightAttribute string
+	RightAttribute string `yaml:"RightAttribute"`
```
Juneezee, line 38: Unlike `github.com/ghodss/yaml` which first converts YAML to JSON then JSON to struct [^1], `gopkg.in/yaml.v2` unmarshals using the field name lowercased as the default key [^2].

[^1]: https://pkg.go.dev/github.com/ghodss/yaml#Unmarshal
[^2]: https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal

cdujeu:
nice spot although I think the dependency is in fact only used by the corresponding test file (not the actual code)

Issue #493, opened 2023-06-20
Error “Pydio is not defined” when run build-core command in gui.ajax folder


HaiyangZhang1994:
_Note: This issue tracker is only used for reporting **qualified bugs**. Use [Pydio Forums](https://forum.pydio.com) for support or feature requests: first-start, install, or config-related troubleshootings will be ignored and closed! Otherwise, please **fill all sections below**._  

**📝 Describe the bug**

When I try to run “pnpm run build-core” under folder gui.ajax and then make all, seems like the project start and get error “error loading core/pydio.min.js?v=8d99e3b8bcc1269270e895514b7d9833:Pydio is not defined” from webpage, does anyone has the same issue?

**⚙️ How-to Reproduce**

1. Use terminal enter gui.ajax's folder
2. Run "pnpm run build-core"
3. Run "make all" under main folder
4. Run "./cells start" under main folder

**🩺 Environment / Setup**

Complete the following information:

Server Versions:
 - Cells Version 4.2.2
 - MariaDB/MySQL version MySQL 8.0.33
 - Server OS: macOS 12.3.1
 - Other dependencies None

Client used for testing: 
 - Browser chrome Version 114.0.5735.133
 - Client OS / mobile device, etc...
 
Additional context:
Here's the screenshot
<img width="510" alt="image" src="https://github.com/pydio/cells/assets/44517160/75b65218-ccad-40d6-972f-e192baaf9882">

<img width="1122" alt="image" src="https://github.com/pydio/cells/assets/44517160/98f6368f-e171-43e1-a6a4-f89ddee701bb">

<img width="1116" alt="image" src="https://github.com/pydio/cells/assets/44517160/078c80a6-ee4d-4d1f-9495-d5f71bfb4b80">

HaiyangZhang1994:
@cdujeu

cdujeu:
hello, can you give more context why you want to re-build ? 
Did you also `pnpm run build-libs` and `pnpm run build-boot` ? 

Maybe first try to `make dev` (will use the local JS instead of embed it inside the binary) to validate your JS changes
Also do not forget to env NODE_ENV=production before rebuilding the JS to make sure it is minified

HaiyangZhang1994:
> hello, can you give more context why you want to re-build ? Did you also `pnpm run build-libs` and `pnpm run build-boot` ?
> 
[...]

PR #532, opened 2024-02-20
webpack5 arrow function arguments compatibility of gui/core module.

chaosue:
after rebuild of gui.ajax core module `pnpm run build-core-prod`, the page will end up with the errors below:

图片

图片
maybe it's the same issue as of #493.     

there're some explanations on https://www.ginkonote.com/users/flo/articles/migration-guide-to-webpack-5@javascript

the cause might be  "arguments objects is not available in arrow function after Webpack 5+",  
the packed `frontend/front-srv/assets/gui.ajax/res/js/core/index.js` is not evalable, so the loading of pydio.min.js by pydio.boot.in.js fails.

### DIFF
--- a/frontend/front-srv/assets/gui.ajax/res/js/core/index.js
+++ b/frontend/front-srv/assets/gui.ajax/res/js/core/index.js
@@ -175,8 +175,7 @@ if(originalRequire) {
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


ghecquet:
CLA not signed and a lot happened since, closing.

PR #536, opened 2024-02-26
fix defunct processes and deregister server before it stops

chaosue:
图片

as shown above, after stopping the master cells process, there might be some sub processes left defunct in random. 
two points in cells may lead to this problem.

1. The `Stop` method in main routine does not wait the `StartAndWait` method to complete waiting its sub process.
2. The `gateway.data` plugin , which uses the minio package, does not fork, so it still serves in the main process, while the minio's exit function will directly end the main process when stopping the pulgin. However the minio's exit function cannot be customized, so there's no way to tweak this behaviour from the minio side at present. Only the forking way may resolve it.

so whether there will be defunct processes depends on who (the main process and its sub processes) exits first in random.

another point, shall we deregister the server to notify the client-ends in advance before stopping the server to prevent unexpected future requests.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -24,16 +24,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/url"
-	"os"
-	"path/filepath"
-	"strings"
-
 	"github.com/manifoldco/promptui"
-	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


ghecquet:
CLA not signed and a lot happened since - closing

PR #554, opened 2024-03-22
fix some i18n missing strings on UI

listeng:
fix some i18n missing strings on UI


cdujeu:
hello, thanks for submitting
but this is not really how it works : we only set the en-us strings, then go through CrowdIn to make the translations. 
We'll merge your suggestion manually!
Thx

listeng:
Thanks for your work. The strings in this commit are not available in CrowdIn, because they are hard-coded in UI. As a result, we cannot translate these strings using CrowdIn.

cdujeu:
i will ping you when they are in CrowdIn

cdujeu:
(fixed in code - soon to appear in CrowdIn)

PR #573, opened 2024-06-22
Corrected a translation error in "Japanese."

soramikan:
"Japanese" in [common/utils/i18n/i18n.go](https://github.com/pydio/cells/blob/a0e30b5f1b5657e2d5b9badd81ad8740c7a38197/common/utils/i18n/i18n.go#L50) for "日本語", not "日本人".
"日本人" means people, but "日本語" means language.


cdujeu:
thanks @soramikan, i manually changed the code

PR #586, opened 2024-10-03
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of length rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW


cdujeu:
manually applying as cla is not signed

cuishuang:
> manually applying as cla is not signed

Where is the cla?

PR #673, opened 2025-10-07
fix(folder-size-cache): use redis cache to store folder sizes

osalama7:
Folder sizes are displayed incorrectly, fixed by using shared redis cache.

### DIFF
--- a/common/storage/sql/index/folder-size-cache.go
+++ b/common/storage/sql/index/folder-size-cache.go
@@ -213,7 +213,7 @@ func (dao *FolderSizeCacheSQL) invalidateMPathHierarchy(ctx context.Context, mpa
 
 // Compute sizes from children files - Does not handle lock, should be
 // used by other functions handling lock
-func (dao *FolderSizeCacheSQL) folderSize(ctx context.Context, node tree.ITreeNode) (err error) {
+func (dao *FolderSizeCacheSQL) folderSize(ctx context.Context, node tree.ITreeNode) error {
 	mpath := node.GetMPath().ToString()
 	ca, err := cache_helper.ResolveCache(ctx, common.CacheTypeShared, folderSizeCacheConfig)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


common/storage/sql/index/folder-size-cache.go
```diff
@@ -53,7 +58,10 @@ func (dao *FolderSizeCacheSQL) GetNodeByMPath(ctx context.Context, path *tree.MP
 	}
 
 	if node != nil && !node.GetNode().IsLeaf() {
-		dao.folderSize(ctx, node)
+		err := dao.folderSize(ctx, node)
+		if err != nil {
+			log.Logger(ctx).Debug("Unable to use cache for folder size", zap.Error(err))
+		}
```
cristianoliveira, line 64: Suggestion, since this is mostly for logging, the call to the logger could be moved to `folderSize` function. The `ctx` is passed to that function too

common/storage/sql/index/folder-size-cache.go
```diff
@@ -141,80 +161,82 @@ func (dao *FolderSizeCacheSQL) GetNodeTree(ctx context.Context, path *tree.MPath
...
 
-	folderSizeLock.Lock()
-	folderSizeCache[mpath] = size
-	folderSizeLock.Unlock()
+	// using a string representation for byte size since redis does not support int64.
+	err = ca.Set(mpath, []byte(strconv.FormatInt(int64(size), 10)))
+	if err != nil {
+		return err
+	}
+	return nil
```
cristianoliveira: Not sure this check is necessary
```suggestion
	return err
```

common/storage/sql/index/folder-size-cache.go
ghecquet: Oops my comment didn't make it through. I've asked him to do that - so that we always return the error to the highest level possible in the stack. That avoids duplicate error logging.

common/storage/sql/index/folder-size-cache.go
osalama7: I removed the err return, just return

PR #685, opened 2025-10-17
chore(dev-ex): adding live reload config

osalama7:


.air.toml
```diff
@@ -0,0 +1,53 @@
...
+  exclude_file = []
+  exclude_regex = ["_test.go"]
+  exclude_unchanged = false
+  follow_symlink = false
+  full_bin = ""
+  include_dir = []
+  include_ext = ["go"]
+  include_file = []
+  kill_delay = "0s"
+  log = "build-errors.log"
```
cristianoliveira: Can it be?
```suggestion
  log = "*.log"
```

cristianoliveira: Ahh this is for the output, nvm

.air.toml
```diff
@@ -0,0 +1,53 @@
...
+testdata_dir = "testdata"
+tmp_dir = "tmp"
+
+[build]
+  root = "."
+  args_bin = ["start"]
+  bin = "./cells"
+  cmd = "make dev"
+  delay = 1000
+  exclude_dir = ["assets", "tmp", "vendor", "testdata"]
```
cristianoliveira, line 11: Does it respect the git ignore? I usually have `.tmp/gitworktrees` ignored globally

.air.toml
```diff
@@ -0,0 +1,53 @@
...
+  exclude_file = []
+  exclude_regex = ["_test.go"]
+  exclude_unchanged = false
+  follow_symlink = false
+  full_bin = ""
+  include_dir = []
```
cristianoliveira: Maybe here, instead of including everything, list the dirs, that would solve the issue with having tmp folders in the repo when working locally (the thing I mentioned about the git worktree)

.air.toml
osalama7: Thats fine, I'll commit it from my machine.

.air.toml
osalama7: Do you mean does it also ignore files in `.gitignore`?
Its not mentioned [explicitly](https://github.com/air-verse/air)

cristianoliveira: Not only that but user's specific ignored files usually in `$HOME/.gitignore`

.air.toml
osalama7: Pushed a commit for this

.air.toml
cristianoliveira: I think the change to explicitly include the watched dirs solve this dbdc1c42a4f2fd207205caa6fc1ed71bd1931911

cristianoliveira (approved):
LGTM! 🎉

cdujeu (approved):
lgtm

cdujeu (approved):

PR #686, opened 2025-10-17
fix(thumbs): WPB-21131 generate thumbs with the exif orientation

cristianoliveira:
Summary:
This commit removes the custom handling of EXIF orientation from the JavaScript components, simplifying the rendering logic. It adds support in Go code for resolving the orientation based on the EXIF tag.

Detailed Changes:
 - scheduler/actions/images/encoding/codec.go:
   - Implemented the use of `imageorient.Decode` to handle image orientation based on EXIF metadata in the image decoding process.
 - scheduler/actions/images/encoding/codec_test.go:
   - Added test to verify orientation handling in the decoding process.
 - scheduler/actions/images/encoding/exif.go:
   - Removed the metadata `image_exif_orientation` injection for new uploads

DEPS
 - go.mod:
   - Added `github.com/disintegration/imageorient` package for handling image orientation based on EXIF metadata.

## How to check

  - First upload an image that contains an orientation in exif (see the attached examples)
  - Upon upload, the thumbnails should be presented in a intuitive orientation
  - (extra check) Inspect the HTML the thumbs should not contain the classes `ort-rotate-x`
  - Look also the preview in Home, they should be displayed correctly
  
Sample:
[orientation.zip](https://github.com/user-attachments/files/22970075/orientation.zip)


scheduler/actions/images/encoding/codec_test.go
```diff
@@ -0,0 +1,123 @@
...
+		So(err, ShouldBeNil)
+		bounds := img.Bounds()
+		So(bounds.Dx(), ShouldEqual, sourceHeight)
+		So(bounds.Dy(), ShouldEqual, sourceWidth)
+		So(approxColor(img.At(0, 0), color.RGBA{R: 255, G: 255, B: 255, A: 255}), ShouldBeTrue)
+		So(approxColor(img.At(0, 1), color.RGBA{A: 255}), ShouldBeTrue)
+	})
+}
+
+func buildJPEGWithOrientation(t *testing.T, orientation uint16) []byte {
```
cristianoliveira: For this test, I think building the image on memory like this makes the test a bit hard to understand, maybe I'll use a fixture image

scheduler/actions/images/encoding/codec.go
```diff
@@ -25,6 +25,7 @@ import (
 	"image/color"
 	"io"
 
+	"github.com/disintegration/imageorient"
 	"github.com/disintegration/imaging"
```
cristianoliveira, line 29: Maybe we can replace this library as a next step

frontend/assets/editor.diaporama/res/js/preview.js
```diff
@@ -65,4 +52,4 @@ export default ({node, containerWidth, ...remainingProps}) => {
         }}
     />);
 
-}
```
cristianoliveira, line 68: We could introduce prettier to enforce removal of the trailing spaces

cristianoliveira:
Looks good!
## Demo

https://github.com/user-attachments/assets/a02126b0-ba49-43f2-989e-ffae69e2f3b2

frontend/assets/editor.diaporama/res/js/editor.js
cdujeu, line 1: I don't agree here : **old** thumbnails will still need the CSS fix!

=> We should let all that code here, and in the backend, just ensure that we do not set this metadata anymore for new ones (EXIF extraction job)

cdujeu (changes requested):
will not be backward compat I think

frontend/assets/editor.diaporama/res/js/editor.js
cristianoliveira: That makes sense

I changed to stop injecting the exif orientation for now on, so the new uploads don't contain it, while the old that contains the metadata, the css rotation is applied

## Demo

https://github.com/user-attachments/assets/2c266942-2651-4424-a06b-107d51a6038f

scheduler/actions/images/encoding/codec.go
```diff
@@ -82,7 +83,8 @@ func NewImageCodec(fileExt string) ImageCodec {
 
 // Decode reads an image from the provided reader
 func (c defaultCodec) Decode(reader io.Reader) (image.Image, error) {
-	return imaging.Decode(reader)
```
cdujeu, line 85: Sorry to bother again :-) 
Does this have any perf implication for formats that do not support exif ? It seems that exif is JPG/TIFF only, maybe we should use the "standard" decode for BMP, PNG, WEBP...?

cdujeu (commented):
Last minor check on formats :-)

scheduler/actions/images/encoding/codec.go
cristianoliveira: There might be a minor performance hit. For our case I believe is better to configure it upon decoder instantiation, to avoid having non-intuitive rotation only for one format type. Something like:
```go
	t.codec = encoding.NewImageCodec(fileFormat, &encoding.CodecOptions{
		EnforceExifOrientation: true,
	})
```

Then in case is not enforced it uses the standard encoding

cristianoliveira: I notice also that there are some img types that simply don't support exif at all, for instance, GIF / BMP. I see the underlining library handle these already

cdujeu (approved):
Ok for me

PR #687, opened 2025-10-21
fix(search): ensure tags filter match exactly one result

osalama7:
- Refactored regexComaTerms to use regex patterns that match tags as whole words in comma-separated lists.
- Prevents partial matches (e.g., filtering for `value` no longer matches `value` AND possibly `values`).
- Added helper functions for clean term extraction and regex construction.

### DIFF
--- a/data/search/dao/mongo/codec.go
+++ b/data/search/dao/mongo/codec.go
@@ -3,6 +3,7 @@ package mongo
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


data/search/dao/mongo/codec.go
```diff
@@ -223,17 +224,23 @@ func (m *Codex) regexTerm(term string) primitive.Regex {
...
-func (m *Codex) regexComaTerms(metaName string, terms []string, not bool) (filters []bson.E) {
+func (m *Codex) regexComaTerms(metaName string, terms []string, not bool, exact_match bool) (filters []bson.E) {
 	for _, part := range terms {
 		tok := strings.TrimSpace(part)
 		if tok == "" {
 			continue
 		}
+		var pattern string
+		if metaName == "meta.usermeta-tags" && exact_match {
+			pattern = "^" + regexp.QuoteMeta(tok) + "$"
```
cdujeu: shall we maybe use QuoteMeta in any cases? (exact match or not)

data/search/dao/mongo/codec.go
```diff
@@ -223,17 +224,23 @@ func (m *Codex) regexTerm(term string) primitive.Regex {
...
 // regexComaTerms replaces multiple values by multiple regexes
-func (m *Codex) regexComaTerms(metaName string, terms []string, not bool) (filters []bson.E) {
+func (m *Codex) regexComaTerms(metaName string, terms []string, not bool, exact_match bool) (filters []bson.E) {
 	for _, part := range terms {
 		tok := strings.TrimSpace(part)
 		if tok == "" {
 			continue
 		}
+		var pattern string
+		if metaName == "meta.usermeta-tags" && exact_match {
```
cdujeu: No we should not check on the metaName here! We already have the info that this meta namespace is a type Tag in the caller (where we call regexComaTerms)

cdujeu (changes requested):

data/search/dao/mongo/codec.go
osalama7: Checked it after using the latest index collection, works fine, removed metaName check

data/search/dao/mongo/codec.go
osalama7: Makes sense, used on the token param

cdujeu (approved):

osalama7:
Ran integration tests locally, all passing ✅

PR #689, opened 2025-10-23
fix(docs-gen): mkdocs compat

c12simple:
**Description**
Adapting the code to generate mkdocs-compatible docs

**Requires**
"github.com/pydio/go/v2" instead of "github.com/pydio/go"


cdujeu (approved):
ok!

PR #693, opened 2025-10-27
fix(ckeditor): Improving content sanitization for ckeditor

osalama7:
- sanitize file content for ckeditor
- handle on change with xss sanitize


cdujeu (approved):
looks good, tested with the special file

PR #694, opened 2025-10-27
fix(deps): WPB-19702 - update pdfjs and react-pdf to latest version

cristianoliveira:
## Description

Upgrading the dependencies `reac-pdf` and `pdf.js` to latest versions.

### Details
 
 - Removed the old vendor folder
 - Introduced webpack copy plugin to collect and place the correct dependency files
 - Instead of vendoring the entire pdf.js, now we only vendor the files that are related to the web viewer
 - Clean up old files
 
 
 Sorry for the big PR, but there wasn't much to split in this change


cristianoliveira:
I spent about an hour trying to figure out why when building to prod it isn't work, my findings:
The component is not even called when built to prod, so there might be something with it being stripped from the build process. In order to debug that I'll have to deep down into the current plugins build configuration and process, which would consume more time. I left a FIXME in the code to continue it in a second moment

I removed the dead code as agreed, like the config for no_script feature flag

cristianoliveira:
-  Fixed the build-prod to avoid watching mode 

Here is the demo:

https://github.com/user-attachments/assets/95a50228-9146-4160-9e75-5e97fd7ed253

cristianoliveira:
Sorry again for the big PR, but here are a list of the relevant changes:

 https://github.com/pydio/cells/blob/fix/WPB-19702-pdfjs-dist-upgrade/frontend/assets/editor.pdfjs/webpack.config.js
 https://github.com/pydio/cells/blob/fix/WPB-19702-pdfjs-dist-upgrade/frontend/assets/editor.pdfjs/res/js/editor.js
 https://github.com/pydio/cells/blob/fix/WPB-19702-pdfjs-dist-upgrade/frontend/assets/editor.pdfjs/res/js/index.js
https://github.com/pydio/cells/blob/2fcc1438e870e460bc670cc62c35f5dd8752f4f1/frontend/assets/editor.pdfjs/res/dist/pdfjs/web/viewer.css#L16-L24

osalama7 (commented):
lgtm

cdujeu (approved):
Go!

cristianoliveira:
Created the follow up task
https://app.clickup.com/t/869azcyrf

PR #696, opened 2025-10-28
tests: configure vitest and react-testing library for unit tests

cristianoliveira:
### Summary

This pull request introduces a basic unit test setup using Vitest for `assets`. It includes improving component accessibility by adding ARIA attributes. It integrates the necessary testing packages and configurations to allow React component unit testing.

### Additions

- Added a unit test for the `LabelWithTip` component using `@testing-library/react` and `vitest`. The test checks the tooltip's visibility on mouse events and the presence of the label.
- Introduced a `vitest.config.js` file to configure Vitest, specifying React support and the use of JSDOM for a browser-like environment during testing.
- Added various testing-related packages such as `@testing-library/react`, `@testing-library/jest-dom`, `@vitejs/plugin-react`, `jsdom`, and `vitest` to `package.json`.

### Updates

- Refactored `LabelWithTip` by renaming it from `.js` to `.jsx`. Improved component accessibility by adding `aria-label` and `role="tooltip"` attributes.
- Updated scripts in `package.json` to include `test` and `test:watch` commands.

### Notes

 - No particular reason I picked this component to add the first test, I saw it was easy enough and has a bit of logic in it.


frontend/assets/gui.ajax/res/js/ui/Components/elements/LabelWithTip.js
```diff
@@ -57,7 +57,12 @@ export default class LabelWithTip extends React.Component {
                 <span onMouseEnter={this.show.bind(this)} onMouseLeave={this.hide.bind(this)} style={style} className={this.props.className}>
                         {label}
                     {this.props.children}
-                    <div label={this.props.tooltip} style={{...tooltipStyle,display:this.state.show?'block':'none'}} className={this.props.tooltipClassName}/>
```
cristianoliveira, line 60: `label` is an invalid HTML property for `div`, I believe we wanted `aria-label`, right?

cristianoliveira:
- Next step, we need to ensure the test files aren't included in the binary ✅

frontend/assets/webpack-commons.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
                 rules: [
                     {
                         test: /\.(js|jsx)$/i,
+                        exclude: /(\.(spec|test)\.jsx?$)/,
```
cristianoliveira: Maybe we can introduce this change in a follow up PR, because we may need to regenerate all bundled files to ensure it's all up to date. WDYT?

cdujeu (approved):

PR #697, opened 2025-10-29
chore/add prettier config

osalama7:
Adding Prettier 

- The config formatting does not change the current spacing convention
- Run prettier per asset to avoid huge pull requests
- use prettier with npx `npx prettier <path> --check` or `npx prettier <path> --write` to fix problems


cristianoliveira (approved):
LGTM

PR #701, opened 2025-10-31
fix(pdfjs): adds missing import css for react-pdf

cristianoliveira:
Fixes the warning and the error with the missing `.hiddenCanvasElement` that causes issues when resizing the window.

See: https://github.com/wojtekmaj/react-pdf#support-for-text-layer


cdujeu (approved):
lgtm, no more black rectangle!

PR #715, opened 2025-11-24
feat(namespaces): optimize name space model changes

osalama7:
- Remove NodeType
- Move prompt options boolean value to namespace root
- Changing json schema type to pbstruct
- Regenerate protobuf and swagger
- Updating access package to set new promptOptions value
- Use new regenerated js sdk path
- Update prod builds


cdujeu:
lgtm @osalama7

PR #718, opened 2025-11-26
feat: WPB21998 - adds WithPresignedURLs flag for `/versions`

cristianoliveira:
## Details

In order to be able to download a given version directly, we need to have a way for the client to use a presigned URL to use with collabora.

### Approach
  - Added the new Flags options
  - When passing the flag it returns the presigned url
  
 ### How to check

Calling this curl with a token generated by `./cells admin user token -u admin -a 10m`  and a valid node. Make sure to version the datasource.
 ```bash
 curl -k 'https://localhost:8080/v2/n/node/2a6b4692-cacc-4773-afcc-5e94ebdccebc/versions' \
  -H "authorization: Bearer $TOKEN" \
  -H 'content-type: application/json' \
  --data-raw '{"SortField": "MTime", "SortDirDesc": true, "Flags":["WithPreSignedURLs"]}'
```
It should return the new property

By accessing the PreSignedGET you should get a version of the node with the changes of that version


cdujeu (approved):
lgtm!

PR #722, opened 2025-11-27
feat(namespace): Json Schema factory

osalama7:
- Add GetNamespaceSchema RPC to UserMetaService and unify field responses to JsonSchemaResponse in cells-idm.proto.
- New endpoint /user-meta/namespace/jsonschema.
- Regenerate swagger entries.
- Implement DAO support to return combined namespace JSON.
- Add gRPC handler for GetNamespaceSchema.
- Change GetFieldSchema to return JsonSchemaResponse.
- Add REST handler to expose GetNamespaceSchema.
- Refactor idm/meta/json_schema.
- Introduce MetaSchemaFactory and JSONSchemaFactory and helper builders.
- Add schema.go wrapper for github.com/google/jsonschema-go/jsonschema (Infer helpers).
- Implement BuildNamespacesJsonSchema, GetMetaSchema, GetJsonSchema, LegacyTypeToLabel, and helper functions.
Update
- Add tests to cover meta/json schema behavior.
- Update `cells-sdk-js` and `ajax.gui` to use latest sdk
- Expose individual ns jsons schemas for onChange field validation
- Exposed unified ns json schema

- Next steps
  - cover more legacy ns defs for star rating and labels
  - (Improve error handling)
  - Integration test

 - Testing changes
Unified schema
```bash
  curl --location 'https://localhost:8080/a/user-meta/namespace/jsonschema' \
--header 'type: string' \
--header 'Authorization: Bearer<AdminToken>'
```
MetaSchema
```bash
curl --location 'https://localhost:8080/a/user-meta/namespace/discovery/field/string' \
--header 'type: string' \
--header 'Authorization: Bearer <AdminToken'
```
Namespace Schema
```bash
curl --location 'https://localhost:8080/a/user-meta/namespace' \
--header 'type: string' \
--header 'Authorization: Bearer <AdminToken'
```


idm/meta/json_schema/json_schema.go
```diff
@@ -0,0 +1,263 @@
...
+			"type":       "object",
+			"properties": map[string]interface{}{},
+		},
+	}
+}
+
+type NamespaceDescriptor struct {
+	Label      string
+	Definition []byte
+}
```
osalama7, line 96: A type definition to specify whats needed from Namespaces to avoid unnecessary imports and too many params.

osalama7:
Overall the changes are good to go to support building JsonSchemas for each namespace for meta data field validation, full json schema for all namespaces for tagging submission, and meta schemas for admin console namespace configuration.
Some JSON Definition values for sample types like multi select are not yet supported but they shall be to ensure full backward compatibility.

idm/meta/dao/sql/ns-sql.go
```diff
@@ -188,3 +191,29 @@ func (s *nsSqlImpl) List(ctx context.Context) (map[string]*idm.UserMetaNamespace
 
 	return res, nil
 }
+
+func (s *nsSqlImpl) GetJS(ctx context.Context) (*structpb.Struct, error) {
```
cristianoliveira: Minor: The method name on a first glance seems to me something related to javascript, shall we use something more explicity? like GetJsonSchema?

idm/meta/dao/sql/ns-sql.go
```diff
@@ -188,3 +191,29 @@ func (s *nsSqlImpl) List(ctx context.Context) (map[string]*idm.UserMetaNamespace
...
+		}
+
+		schema, err := json_schema.BuildNamespacesJsonSchema(nss)
+		if err != nil {
+			return nil, err
+		}
+		return schema, nil
+	} else {
+		return nil, nil
+	}
```
cristianoliveira: When using a guard clause, leave the success statement for later. So then you can avoid the complex branches and it's much easier to read. You just look to the end and you see the success clause.

Reference https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html
```suggestion
	if len(mm) = 0 {
    return nil, nil
	}

  nss := make([]json_schema.NamespaceDescriptor, 0, len(mm))
  for _, m := range mm {
    nss = append(nss, json_schema.NamespaceDescriptor{
      Label:      m.Label,
      Definition: m.Definition,
    })
  }

  schema, err := json_schema.BuildNamespacesJsonSchema(nss)
  if err != nil {
    return nil, err
  }

  return schema, nil
```

idm/meta/json_schema/json_schema.go
```diff
@@ -0,0 +1,242 @@
...
+func withDateTimeSchema() map[string]interface{} {
+	s, _ := Infer[string](nil)
+	prop := marshalSchema(s)
+	prop["format"] = "date-time"
+	return prop
+}
+
+func LegacyTypeToLabel(d []byte) string {
+
+	var tv string
```
cristianoliveira, line 231: What is tv? 
Minor: Variable should have meaningful names. For this case in specific should be fine since the context is very small but it doesn't hurt trying to stick with meaningful names

idm/meta/json_schema/json_schema.go
```diff
@@ -0,0 +1,242 @@
...
+		return map[string]interface{}{
+			"minLength": min,
+		}
+	}
+	return map[string]interface{}{
+		"minimum": min,
+	}
+}
+
+func withMax(max int, t string) map[string]interface{} {
```
cristianoliveira, line 186: t string? what is the meaning of t?

idm/meta/dao/sql/ns-sql.go
osalama7: I planned to revisit this part soon, but this makes sense.

common/proto/rest/README.md
```diff
@@ -0,0 +1,60 @@
+# Regenerate JavaScript SDK (quick reference)
+{wip}
```
cristianoliveira, line 2: WIP?

idm/meta/dao/sql/ns-sql.go
osalama7: I tried to keep the name as short as the convention, List, Add, GetJS is short of GetJsonSchema

common/proto/rest/README.md
cristianoliveira: Maybe we should move the generation of the sdk to the repo itself, similarly to what was made for sdk-ts and sdk-swift

idm/meta/json_schema/json_schema.go
osalama7: type value, JsonDefinition looks like this `{"type": "string"}` we need that value to build Schemas on the fly

common/proto/rest/README.md
osalama7: Needs a review perhaps with Charles before adding it into the main Readme.

idm/meta/dao/sql/ns-sql.go
cristianoliveira: Still it's confusing, that's was my understanding of this method, at a first glance

idm/meta/json_schema/json_schema.go
osalama7: It's the type string  = `string` which is the legacy json def. type

idm/meta/dao/sql/ns-sql.go
ghecquet: List, Add are short because they're part of the CRUD for the service. In this case it's better to be explicit

osalama7: Just discussed it with Cristian to try to find a better name in this context, I'll share a suggestion of the new name with the team.

cristianoliveira (approved):

PR #724, opened 2025-12-03
fix(namespace)-WPB-21908: Fixes for json schema persistence on creation

osalama7:
- Remove hard coded meta schema properties to allow values input
- Replace factory generation on ns sql DAO As method with persisted JSON
- Helper function to cast json to pbstruct

### DIFF
--- a/idm/meta/dao/sql/ns-sql.go
+++ b/idm/meta/dao/sql/ns-sql.go
@@ -23,6 +23,7 @@ package sql
 import (
 	"context"
 
+	"google.golang.org/protobuf/types/known/structpb"
 	"gorm.io/datatypes"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
@@ -35,7 +36,6 @@ import (
 	"github.com/pydio/cells/v5/common/storage/sql/resources"
 	"github.com/pydio/cells/v5/common/telemetry/log"
 	"github.com/pydio/cells/v5/idm/meta"
-	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/pydio/cells/v5/idm/meta/json_schema"
 )
--- a/idm/meta/dao/sql/ns-sql.go
+++ b/idm/meta/dao/sql/ns-sql.go
@@ -28,7 +28,6 @@ import (
 	"gorm.io/gorm/clause"
 	"gorm.io/gorm/schema"
 
-	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/pydio/cells/v5/common/errors"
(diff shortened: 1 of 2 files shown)


idm/meta/dao/sql/ns-sql.go
```diff
@@ -28,14 +28,14 @@ import (
...
 
-	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/pydio/cells/v5/common/errors"
 	"github.com/pydio/cells/v5/common/proto/idm"
 	"github.com/pydio/cells/v5/common/proto/service"
 	"github.com/pydio/cells/v5/common/storage/sql"
 	"github.com/pydio/cells/v5/common/storage/sql/resources"
 	"github.com/pydio/cells/v5/common/telemetry/log"
 	"github.com/pydio/cells/v5/idm/meta"
+	"google.golang.org/protobuf/types/known/structpb"
```
cdujeu: imports ordering do you run goimports-reviser ? 
Should group imports by stdlib, external imports, "internal" imports (a.k.a cells)

idm/meta/json_schema/json_schema.go
```diff
@@ -129,46 +154,46 @@ func BuildNamespacesJsonSchema(ns []NamespaceDescriptor) (*structpb.Struct, erro
 	return toProtoStruct(root), nil
 }
 
-func (f *JSONSchemaFactory) BuildJsonSchema(label string) *structpb.Struct {
+func (f *JSONSchemaFactory) BuildJsonSchema(label string) ([]byte, *structpb.Struct, error) {
 	props := f.root["properties"].(map[string]interface{})
-
+	f.root["$id"] = fmt.Sprintf("https://schemas.pydio.com/%s", label)
```
cdujeu: do we have to make a decision on that URL (the host part)? Will it be permanently used? 
If so at least put it as a const somewhere (or even better a parameter of your Factory?)

cdujeu (commented):
all good appart from my minor comments

idm/meta/json_schema/json_schema.go
osalama7: We don't have to keep the Ids or use schema refs for now, I intended to leave it the schema strcuture until we needed it, I added it there as `wip` to validate schema merging on the client side. I'll remove it in the next PR.

idm/meta/dao/sql/ns-sql.go
osalama7: I have `go imports-reviser` installed but changed settings in VScode on save at some point. 
Will fix this in my next commit. added `TODO`

PR #731, opened 2025-12-09
fix(api): accept any name for namespace schema property

cristianoliveira:
This commit changes the JSON schema building logic to correctly handle namespace properties regardless of their key name in the source schema.

Detailed Changes:
 - Implementation:
   - Changed json_schema.go to iterate over properties map instead of assuming a specific key name, allowing any property name in the namespace schema.
   - Added validation to reject schemas with more than one property definition for a namespace.
   - Changed the title assignment for textarea fields to use the actual field name instead of a hardcoded string.
 - Tests:
   - Updated json_schema_test.go to match the new property structure in test expectations.


osalama7 (approved):
lgtm

PR #738, opened 2025-12-15
WPB22258 fix: handle empty metadata

cristianoliveira:
This commit fixes an issue where empty user meta data was incorrectly processed during uploads, causing sessions to fail.

Detailed Changes:
 - Implementation:
   - In UploadItem.js, changed the condition to check if user meta map size is greater than zero before processing.
   - In DropUploader.js, changed the metaDismiss method to convert the data object to a Map before passing to session.

### Demo

https://github.com/user-attachments/assets/d3047c42-22c1-4403-a64e-0fd3f7d80409


osalama7 (approved):
lgtm

PR #743, opened 2025-12-16
fix(jsonschema): remove format field from string schema

cristianoliveira:
This commit removes the format field from the string schema in JSON Schema factory.

Detailed Changes:
 - Implementation:
   - Removed the format field from the string schema in json_schema.go.
   - Updated test calls in json_schema_test.go to include an empty string parameter for BuildJsonSchema.

(cherry picked from commit d2cc2043f8b41278cfa685eeea1ba7e57cc9d4b6)


osalama7 (approved):
lgtm

PR #753, opened 2025-12-18
feat(metadata): fix uploader glitches:

cdujeu:
- drop on homepage opens the prompt,
- do not append virtual node in the list before metadata is filled
- repair folders creation,
- big folders can take some time to process before closing metadata panel, show a transient status


cristianoliveira (approved):
lgtm

PR #756, opened 2025-12-18
fix(metadata): ignore empty fields in global validation

cristianoliveira:
This commit updates the validation logic to ignore empty fields during global validation, preventing false errors for required fields that are intentionally left blank.

Detailed Changes:
 - Implementation:
   - Changed the test schema to add a new required datetime field and updated the required fields list.
   - Updated test fixtures to replace 'dates' namespace with 'usermeta-datetime' for consistency.
   - Modified validation tests to use the updated namespace and added a new test case for ignoring empty fields in global validation.
 - Tests:
   - Added a new test case that verifies empty fields are ignored during global validation, ensuring validation only fails for actual invalid data.
   
  ##Demo
  
https://github.com/user-attachments/assets/7797b369-5566-4e62-b359-4cfea1b80ea1


osalama7 (approved):
lgtm

PR #760, opened 2025-12-29
fix(cors): expand CORS support with defaults for certain routes and ability to override them at per-site / per-route level in Sites definition.

cdujeu:
Expand CORS : 
 - Initially fix missing CORS at Caddy level for OIDC endpoint, by declaring them directly in the "oidc" route. 
 - Prepare caddyfile for supporting CORS at various level (with override) : global (ENV), site-wide (via ProxyConfig), route-wide (via defaults added above, or via ProxyConfig).


ghecquet (approved):
Just as a tiny general comment, can we please keep acronyms uppercase and standardise them across the code ? We have a mixture of CORS, Cors. I think it should always be uppercase apart from the package name

PR #762, opened 2026-01-06
feat(namespaces): Add metadata prompt feature flag

osalama7:
- Add global param for metadata prompt ff 
- Hide prompt options and required switches with Prompt feature flag


frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -402,23 +404,16 @@ class MetaNamespace extends React.Component {
...
                     {...ModernStyles.toggleFieldV2}
-                />
-                <Toggle
-                    label={m('toggle.list-visibility')}
-                    disabled={readonly}
-                    labelPosition={"left"}
-                    toggled={!this.getHideValue()}
-                    onToggle={(e, v) => { this.setHideValue(!v) }}
-                    {...ModernStyles.toggleFieldV2}
-                />
```
osalama7, line 421: This toggle switch was duplicated at some point, and removed in this commit.

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -280,6 +280,8 @@ class MetaNamespace extends React.Component {
         const { create, namespaces, pydio, readonly, muiTheme } = this.props;
         const { namespace, m, PoliciesBuilder, metaModule } = this.state;
         const ModernStyles = ThemedModernStyles(muiTheme)
+        const configs = this.props.pydio.getPluginConfigs('access.settings');
+        const USERMETA_PROMPT_FF = configs.get('USERMETA_NAMESPACE_PROMPT') || false;
```
cristianoliveira:
```suggestion
        const USERMETA_PROMPT_FF = configs.get('USERMETA_NAMESPACE_PROMPT');
```
Small fix, I don't think that is necessary. That would be the same as `false || false` 😄
In case the goal was to ensure its type is boolean then I'd suggest using `Boolean(val)`

cristianoliveira (approved):
Besides the small fix, LGTM

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
osalama7: Committed a new change with a Boolean constructor, Trying to keep the flag value false if config map returns undefined.

PR #767, opened 2026-01-16
fix(namespace): Fix namespace json schema builder to handle schemas

osalama7:
- Handle schemas with empty properties


cdujeu (approved):
plz merge!

PR #769, opened 2026-01-21
feat/WPB-22915(meta): add entity-value storage layer for user metadata

osalama7:
Add new entity-value DAO to support structured metadata with entities and their values, enabling better organization of user metadata through relational associations.

- Add MetaEntity and EntityValue proto messages
- Implement EntityValueDAO with CRUD and linking operations
- Add test coverage for entity operations and value management
- Integrate with main meta DAO and migration support


idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,319 @@
...
+	Label      string `gorm:"column:label;type:varchar(100);notNull;uniqueIndex:idx_entity_label"`
+	EntityUUID string `gorm:"column:entity_uuid;type:varchar(255);uniqueIndex:idx_entity_label"`
+}
+
+func (*MetaEntityValue) TableName(namer schema.Namer) string {
+	return namer.TableName("meta_entitiy_values")
+}
+
+// MetaValueEntity junction table model
+type MetaValueEntity struct {
```
cdujeu: i _think_ (to be verified) that there is a more "gorm-like" way of defining n-to-n relationships and proper foreignkeys than managing the join table ourselves. i might be wrong though, can you double check with @ghecquet ?

osalama7: True, I shared a link with him earlier in Chat.

osalama7: Added Gorm Relations

ghecquet: you're not actually setting the foreign key. In case of a many to many relationship you can check the documentation here : https://gorm.io/docs/many_to_many.html

There are also examples in the code here : https://github.com/pydio/cells/blob/3b2ba547387095b62e7f43c6dfb637749d0b6a4d/common/proto/idm/cells-idm.proto#L625C14-L625C27

also, be careful with the naming, (meta_entity, meta_entity_value, meta_value_entity). For a relationship table, merge the two table names and end with rel => meta_entity_value_rel

osalama7:
To fix issues with buf generation mentioned earlier, resolved by deleted the local `/bin` folder, then re-run `buf.sh` to reinstall internally maintained binaries in [this path](https://github.com/pydio/cells/tree/v5-dev/cmd/protoc-gen-go-client-stub)

> Having a small issue with proto buf generation `bash buf.sh idm`  generates undesired changes in _grpc.stub.pb.go

idm/meta/dao/sql/ev-sql.go
osalama7: I updated the relations, and tested migrations and relation constraints.
I had a few hiccups with migrations for the following names

```
idm_usr_meta_entities
idm_usr_meta_values -> this caused issues with migrations (would be ideal but the meta prefix makes it confusing
idm_usr_meta_values_rel -> this worked but needed to be removed from the migrations, gorm generations rel tables automatically. 
```

common/proto/idm/cells-idm.proto
```diff
@@ -490,6 +490,21 @@ message UserMetaNamespace{
...
+    string Uuid = 1;
+    string Label = 2;
+    string Description = 3;
+    string LabelI18N = 4;
+}
+
+// EntityValue represents a value for an entity (e.g., "Engineering", "Sales")
+message EntityValue {
+    string Uuid = 1;
+    string Label = 2;
```
cdujeu, line 503: Shall we directly have a LabelI18N for EntityValue as well? 

BTW what about using I18nLabelsJSON (plural & json) to make this field directly more descriptive? I think at least the plural could be nice.

idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,251 @@
...
+	return namer.TableName("meta_entity_values")
+}
+
+type MetaValuesRel struct {
+	MetaUUID   string `gorm:"primaryKey;column:meta_uuid;type:varchar(255);notNull;index:idx_meta_evalue,composite:meta_evalue"`
+	EValueUUID string `gorm:"primaryKey;column:e_value_uuid;type:varchar(255);notNull;index:idx_meta_evalue,composite:meta_evalue"`
+}
+
+func (*MetaValuesRel) TableName(namer schema.Namer) string {
+	return "idm_usr_meta_values_rel"
```
cdujeu: normally the prefix "idm_usr_" should be added by the Storage abstraction, so it should not appear here

idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func NewEntityValueDAO(db *gorm.DB) meta.EntityValueDAO {
+	return &tagsSqlImpl{
+		Abstract: sql.NewAbstract(db),
+		DAO:      resources.NewDAO(db),
+	}
+}
+
+type tagsSqlImpl struct {
```
cdujeu: i would **not** use the "tags" word here, as tags is always related to TagsCloud values, where here we just want to manage entities & values

idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (u *Entities) FromEntity(res *idm.MetaEntity) *Entities {
+	u.UUID = res.Uuid
+	if u.UUID == "" {
+		u.UUID = uuid.New().String()
+	}
+	u.Label = res.Label
+	u.Description = res.Description
+	if res.LabelI18N != "" {
+		json := datatypes.JSON(res.LabelI18N)
```
cdujeu: does this fail if it's not JSON in input?

idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	return u
+}
+
+func (s *tagsSqlImpl) CreateEntity(ctx context.Context, entity *idm.MetaEntity) (*idm.MetaEntity, error) {
+	res := (&Entities{}).FromEntity(entity)
+	tx := s.Session(ctx).Create(res)
+	if tx.Error != nil {
+		if errors.Is(tx.Error, gorm.ErrDuplicatedKey) {
```
cdujeu: i don't see any differences if its a duplicate or not in the end

common/proto/idm/cells-idm.proto
osalama7: I agree on the naming, and having entity values translated as well, additionally I'll confirm with Greg using a json (`longblob`) type for these columns (JSON was suggested to maintain limits on the implementation scope) if we plan to expose it later we can possibly drop it for now and extend when we expose Entity & Values import.

idm/meta/dao/sql/ev-sql.go
osalama7: True, it was a work around due to the fact the relation table was created by migration from the `many2many` `rel` gorm tag 
To be checked with Greg.

idm/meta/dao/sql/ev-sql.go
osalama7: Had an assertion for `gormDuplicate` error in earlier version which had index on Label
But the index is removed will remove the duplicate entity handling

idm/meta/dao/sql/ev-sql.go
osalama7: No, its unit tested as well.

idm/meta/dao/sql/ev-sql.go
osalama7: We need to use the dynamic table name setter, TBL

common/proto/idm/cells-idm.proto
osalama7: Discussed with Greg to refine the i18n labels in the next stage and remove it from this PR.

idm/meta/dao/sql/ev-sql.go
osalama7: fixed by using the table namer function, Lets observe how migration runs on engineer's machines in the next days to ensure no additional tables are created.

idm/meta/dao/sql/ev-sql.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return createdEntities, nil
+}
+
+func (s *evSqlImpl) GetEntity(ctx context.Context, entityUuid string) (*idm.MetaEntity, error) {
+	var model Entities
+	tx := s.Session(ctx).Where(&Entities{UUID: entityUuid}).First(&model)
+	if tx.Error != nil {
+		if errors.Is(tx.Error, gorm.ErrRecordNotFound) {
+			return nil, nil
```
cristianoliveira, line 145: Hmmm not sure about return nil here, maybe a EmptyError and then check outside. Now the one using this code has to check for 2 nils one for the error and one for the value. Not sure

cristianoliveira: If this is a pattern used everywhere in the codebase, feel free to ignore

cristianoliveira (approved):

idm/meta/dao/sql/ev-sql.go
osalama7: Not really, left over from a previous version of the DAO, its removed in a new commit

PR #780, opened 2026-02-02
feat(namespace): Entity values integration

osalama7:
## General note:
The rest handler logic is done as a work around for entity & values creation using json defintion to maintain backward compatiability and meet the initial goal of this concept, as the logic gets more complicated to handle json parsing and entity & value creation the surface area of errors is increased in this layer for these new namespace types wrapped in If statements. The plan is to add new endpoints to manage these resources.

- add new proto defintions for the new services
- generate gRPC bufs
- Add dao handlers
- expose grpc methods via meta client
- extend rest handler methods update and list namespaces
- add json parsing helpers
- move constants
- extend  json_schema types handler
- update test
- use new mantine Tag cloud input component on admin meta create wizard
- add i18n keys

### DIFF
--- a/common/proto/idm/cells-idm.proto
+++ b/common/proto/idm/cells-idm.proto
@@ -448,7 +448,7 @@ service UserMetaService {
     rpc GetFieldSchema(GetFieldSchemaRequest) returns (JsonSchemaResponse) {}
     rpc GetNamespaceSchema(GetNamespaceSchemaRequest) returns (JsonSchemaResponse) {}
     rpc GetEntityValues(GetMetaEntityValuesRequest) returns (MetaEntityValueResponse) {}
-    rpc DeleteEntityValuesData(GetMetaEntityValuesRequest) returns (DeleteEntityValuesResponse) {}
+    rpc DeleteEntity(GetMetaEntityValuesRequest) returns (DeleteEntityValuesResponse) {}
     rpc CreateEntity(CreateEntityRequest) returns (CreateEntityResponse) {}
     rpc CreateEntityValues(CreateEntityValueRequest) returns (CreateEntityValueResponse) {}
 }
--- a/common/proto/idm/cells-idm.proto
+++ b/common/proto/idm/cells-idm.proto
@@ -447,6 +447,38 @@ service UserMetaService {
     rpc ListUserMetaNamespace(ListUserMetaNamespaceRequest) returns (stream ListUserMetaNamespaceResponse){}
     rpc GetFieldSchema(GetFieldSchemaRequest) returns (JsonSchemaResponse) {}
     rpc GetNamespaceSchema(GetNamespaceSchemaRequest) returns (JsonSchemaResponse) {}
+    rpc GetEntityValues(GetMetaEntityValuesRequest) returns (MetaEntityValueResponse) {}
+    rpc DeleteEntityValues(GetMetaEntityValuesRequest) returns (DeleteEntityValuesResponse) {}
+    rpc CreateEntity(CreateEntityRequest) returns (CreateEntityResponse) {}
+    rpc CreateEntityValues(CreateEntityValueRequest) returns (CreateEntityValueResponse) {}
+}
+
+message CreateEntityValueRequest{
+    repeated EntityValue EntityValue = 1;
+}
+
+message CreateEntityValueResponse{
+    repeated EntityValue EntityValue = 1;
+}
+
+message CreateEntityRequest{
+    MetaEntity Entity = 1;
+}
+
+message CreateEntityResponse{
+    MetaEntity Entity = 1;
+}
+
+message DeleteEntityValuesResponse{
+    int64 RowsDeleted = 1;
+}
+
+message GetMetaEntityValuesRequest{
+    string EntityUuid = 1;
+}
+
+message MetaEntityValueResponse{
+    repeated EntityValue EntityValue = 1;
 }
 
 // Piece of metadata attached to a node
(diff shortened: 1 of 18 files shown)


idm/meta/rest/handler.go
```diff
@@ -161,15 +182,104 @@ func (s *UserMetaHandler) UpdateUserMetaNamespace(req *restful.Request, rsp *res
 		if _, e := ns.UnmarshallDefinition(); e != nil {
 			return errors.WithMessagef(errors.UnmarshalError, "invalid json definition for namespace: %s, %v", ns.Namespace, e)
 		}
+		if input.Operation == idm.UpdateUserMetaNamespaceRequest_PUT {
+			if ns.FieldType != "" {
+				var def jsonDefinition
+				if err := json.Unmarshal([]byte(ns.JsonDefinition), &def); err != nil {
+					return errors.WithMessagef(errors.UnmarshalError, "unable to parse json definition to check existing entity_id for namespace %s: %v", ns.Namespace, err)
+				}
+				if def.Entity.EntityID == "" {
```
osalama7: This implicit link between namespace and entity using JSONDefinition is a bit of a code smell.
One way to solve it is adding a relation between Namespace and Entity

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -25,6 +25,7 @@ import PropTypes from 'prop-types';
 import Pydio from 'pydio'
 import { Dialog, FlatButton, Toggle } from 'material-ui'
 import { muiThemeable } from 'material-ui/styles'
+import { MantineProvider } from '@mantine/core';
```
cdujeu, line 28: check w. Cristian, we expose our own PydioMantineProvider from the "hoc" library. Would be better for centralized theming

idm/meta/dao/sql/ev-sql.go
```diff
@@ -238,3 +266,31 @@ func (s *evSqlImpl) GetMetaEntityValues(ctx context.Context, metaUuid string) ([
 
 	return values, nil
 }
+
+func (s *evSqlImpl) DeleteEntityValues(ctx context.Context, entityID string) (*idm.DeleteEntityValuesResponse, error) {
```
cdujeu: not sure of the name here - should it be DeleteEntity?

idm/meta/rest/handler.go
```diff
@@ -41,6 +43,25 @@ import (
 	"github.com/pydio/cells/v5/idm/meta"
 )
 
+var ns_with_ev = []string{"tag_cloud", "choice", "auto_complete"}
+
+type jsonDefinition struct {
```
cdujeu: We already have the JsonDef schema somewhere near the Namespace protobuf (and a method on this protobuf to load the unmarshalled version)

idm/meta/rest/handler.go
```diff
@@ -161,15 +182,104 @@ func (s *UserMetaHandler) UpdateUserMetaNamespace(req *restful.Request, rsp *res
 		if _, e := ns.UnmarshallDefinition(); e != nil {
 			return errors.WithMessagef(errors.UnmarshalError, "invalid json definition for namespace: %s, %v", ns.Namespace, e)
 		}
+		if input.Operation == idm.UpdateUserMetaNamespaceRequest_PUT {
+			if ns.FieldType != "" {
+				var def jsonDefinition
+				if err := json.Unmarshal([]byte(ns.JsonDefinition), &def); err != nil {
```
cdujeu: See previous, look in proto/idm => namespace

idm/meta/rest/handler.go
cdujeu: not a big problem for now, but to be discussed indeed

idm/meta/rest/handler.go
```diff
@@ -306,3 +436,47 @@ func (s *UserMetaHandler) PoliciesForMeta(ctx context.Context, resourceId string
 
 	return
 }
+
```
cdujeu: same here, i'd rather have that centralized, see proto/idm/meta-ns.go

common/proto/idm/cells-idm.proto
```diff
@@ -447,6 +447,38 @@ service UserMetaService {
...
+
+message CreateEntityResponse{
+    MetaEntity Entity = 1;
+}
+
+message DeleteEntityValuesResponse{
+    int64 RowsDeleted = 1;
+}
+
+message GetMetaEntityValuesRequest{
```
cdujeu, line 476: I think we should support filtering (and maybe offset/limit) already here (even if not yet exposed via rest)

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
osalama7: We paired on this actually, I tried PydioMantineProvider in various options, MetaBoard, AdminBoard and the top level of this plugin and it was always unresolved, will double check tomorrow.

idm/meta/rest/handler.go
osalama7: Will double check.

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
cdujeu: okay nw - park it for later !

idm/meta/dao/sql/ev-sql.go
osalama7: We need to be deleting everything by entityUuid in one function, to keep the integration scope minimal, but it can be revised or a new function that executes all 3 queries in one transaction.

idm/meta/dao/sql/ev-sql.go
```diff
@@ -238,3 +266,32 @@ func (s *evSqlImpl) GetMetaEntityValues(ctx context.Context, metaUuid string) ([
 
 	return values, nil
 }
+
+func (s *evSqlImpl) DeleteEntityValuesData(ctx context.Context, entityID string) (*idm.DeleteEntityValuesResponse, error) {
```
cdujeu: DeleteEntity(ctx, entityID)

cdujeu (approved):

PR #791, opened 2026-02-06
chore: build frontend

cristianoliveira:


osalama7 (approved):

frontend/assets/pnpm-lock.yaml
osalama7, line 1: Any reason why we have lock changes without package.json?

cristianoliveira: Good question. I'm not sure if pnpm was not synched

PR #795, opened 2026-02-09
feat: Adding enforce defaults UX and additional stability fixes

osalama7:
- Clean up types on type editor and group types
- Fix Validation fields logic
- display entity values for tag_cloud and auto_complete in the values field
- Fix Field options display
- Complete Enforce defaults UX

### DIFF
--- a/frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
+++ b/frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
@@ -34,6 +34,18 @@ const { ModernTextField, ModernAutoComplete, ThemedModernStyles } = Pydio.requir
 import FuncUtils from 'pydio/util/func'
 import ResourcesManager from 'pydio/http/resources-manager'
 
+const TYPES_WITHOUT_DEFAULTS = [
+    'tags', 
+    'tag_cloud', 
+    'auto_complete', 
+    'stars_rate', 
+    'css_label', 
+    'json', 
+    'boolean',
+    'choice',
+    'date',
+];
+
 function getGroupValue(namespace) {
     try {
         return JSON.parse(namespace.JsonDefinition).groupName || ""
@@ -138,6 +150,7 @@ class MetaNamespace extends React.Component {
             m: (id) => props.pydio.MessageHash['ajxp_admin.metadata.' + id] || id,
             selectorNewKey: '',
             selectorNewValue: '',
+            namespaceValues: [],
             PoliciesBuilder: MetaPoliciesBuilder
         };
         this.fieldOptionsRef = React.createRef();
@@ -168,7 +181,21 @@ class MetaNamespace extends React.Component {
         if (create && namespaces.length) {
             newNS.Order = namespaces.map(ns => ns.Order || 0).reduce((a, c) => Math.max(a, c), 0) + 1;
         }
-        this.setState({ namespace: newNS });
+        this.setState({ namespace: newNS }, () => {
+            if (props.open && newNS.Namespace && newNS.Namespace !== 'usermeta-') {
+                try {
+                    Metadata.listTags(newNS.Namespace).then(values => {
+                        const merged = values.join(',');
+
+                        this.setState({ namespaceValues: merged });
+                    }).catch(e => {
+                        console.error('Error loading namespace values:', e);
+                    });
+                } catch (e) {
+                    console.error('Error parsing JsonDefinition:', e);
+                }
+            }
+        });
     }
 
     save() {
@@ -446,7 +473,7 @@ class MetaNamespace extends React.Component {
                         <TagsCloudInput
                             label={m('tag_cloud.entity')}
                             disabled={false}
-                            value={this.getCommaSeparatedEntities()}
+                            value={this.state.namespaceValues ? this.state.namespaceValues : ''}
                             onCommitChange={(v) => this.setEntityValues(v)} 
                             hintText={m('tag_cloud.entity.hint')}
                         />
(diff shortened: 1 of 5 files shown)


frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -168,7 +181,21 @@ class MetaNamespace extends React.Component {
...
+
+                        this.setState({ namespaceValues: merged });
+                    }).catch(e => {
+                        console.error('Error loading namespace values:', e);
+                    });
+                } catch (e) {
+                    console.error('Error parsing JsonDefinition:', e);
+                }
+            }
+        });
```
cristianoliveira, line 198: This is an anti-pattern, fetching more state on a callback, that can cause us headaches. Move the fetching to `componentDidUpdate` like
```js
async componentDidUpdate(prevProps, prevState) {
  const prevNamespace = prevState.namespace?.Namespace;
  const nextNamespace = this.state.namespace?.Namespace;

  if (prevNamespace === nextNamespace) return;

  if (!this.props.open) return;
  if (!nextNamespace) return;
  if (nextNamespace === 'usermeta-' /* or startsWith */) return;

  try {
    const values = await Metadata.listTags(nextNamespace);
    // Guard against races: only update if namespace is still current
    if (this.state.namespace?.Namespace !== nextNamespace) return;

    this.setState({ namespaceValues: values.join(',') });
  } catch (e) {
    console.error('Error loading namespace values:', e);
  }
}
```

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -168,7 +181,21 @@ class MetaNamespace extends React.Component {
...
+                try {
+                    Metadata.listTags(newNS.Namespace).then(values => {
+                        const merged = values.join(',');
+
+                        this.setState({ namespaceValues: merged });
+                    }).catch(e => {
+                        console.error('Error loading namespace values:', e);
+                    });
+                } catch (e) {
+                    console.error('Error parsing JsonDefinition:', e);
```
cristianoliveira, line 195: Catch like this I believe is better to just throw, it gives more context to what caused it, this is effectively suppressing the error and hiding it under the "rug" that is the console log 😅. Or let handle it accordingly

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -477,15 +504,32 @@ class MetaNamespace extends React.Component {
...
+                        onToggle={() => this.toggleRequired(namespace)}
+                        {...ModernStyles.toggleFieldV2}
+                    />
+                    {!TYPES_WITHOUT_DEFAULTS.includes(namespace.FieldType) &&
+                        <Toggle
+                            label={"EnforceDefaults"}
+                            labelPosition={"left"}
+                            toggled={namespace?.EnforceDefault}
+                            onToggle={(e, v) => {
+                                namespace.EnforceDefault = v;
```
cristianoliveira, line 525: Shoudln't this be a state? Or if you want to hold this in memory without causing rerendering use a ref. Is this necessary?

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespaceFieldOptions.js
```diff
@@ -32,199 +31,282 @@ import validator from '@rjsf/validator-ajv8';
...
+        NumberInput: {
+            styles: () => ({
+                rightSection: {
+                    display: 'none',
+                },
+                controls: {
+                    display: 'none',
+                },
+                input: {
+                    paddingRight: '0 !important',
```
cristianoliveira, line 50: !important 🙈

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespaceFieldOptions.js
```diff
@@ -32,199 +31,282 @@ import validator from '@rjsf/validator-ajv8';
...
-  useEffect(() => {
-    if (!hasValidNs) return;
-    
-    const nsProperties = ns.JsonSchema?.properties?.[ns.Namespace] || {};
-    const isRequired = ns.JsonSchema?.required?.includes(ns.Namespace) || false;
-    
-    const initialFormData = {
-        ...nsProperties,
-        required: isRequired
+    const customValidate = (formData, errors) => {
```
cristianoliveira, line 99: This here is being recreated every render, ideally we shouldn't rely on closures to hold data `formData` is from the closure, but rather pass as arguments to the function. Please also memoise it with useCallback

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespaceFieldOptions.js
```diff
@@ -32,199 +31,282 @@ import validator from '@rjsf/validator-ajv8';
...
+        const nsProperties = ns.JsonSchema?.properties?.[ns.Namespace] || {};
+        const isRequired = ns.JsonSchema?.required?.includes(ns.Namespace) || false;
+
+        const initialFormData = {
+            ...nsProperties,
+            required: isRequired
+        };
+
+        setFormData(initialFormData);
+    }, [hasValidNs, ns]);
```
cristianoliveira, line 186: I think here it must be more granular, this will rerun every little change in ns. We only use ns.JsonSchema and ns.Namespace

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespaceFieldOptions.js
```diff
@@ -32,199 +31,282 @@ import validator from '@rjsf/validator-ajv8';
...
+
+            try {
+                ajv.compile(patchedSchema);
+            } catch (e) {
+                console.error("Invalid patched schema:", e);
+            }
+
+            return IdmUserMetaNamespace.constructFromObject({
+                ...ns,
+                JsonSchema: ns.JsonSchema,
```
cristianoliveira, line 250: Shouldn't `patchedSchema` be returned here? Why do we need this `getUpdatedNamespace` again?

frontend/assets/meta.user/res/js/TypeEditor.js
```diff
@@ -106,6 +113,26 @@ class TypeEditor extends React.Component {
...
+                        <MenuItem 
+                            key={typeKey} 
+                            value={typeKey} 
+                            primaryText={m('type.' + typeKey) || metaTypes[typeKey]}
+                        />
+                    ))
+            );
+            return items;
+        }, []);
+    }
```
cristianoliveira, line 134: This here deserves test. We can move to a helper function and test it independently

frontend/assets/access.settings/res/js/AdminWorkspaces/model/Metadata.js
```diff
@@ -80,7 +80,26 @@ class Metadata {
...
+                } else {
+                    resolve([]);
+                }
+            }).catch(e => {
+                resolve([])
+            })
+
+        });
+
+    }
```
cristianoliveira, line 102: This looks over complicated
```suggestion
 /**
 * @param {string} namespace
 * @returns {Promise<Array>}
 */
static async listTags(namespace) {
  try {
    const response = await Metadata.api.listUserMetaTags(namespace);
    return response?.Tags ?? [];
  } catch (e) {
    console.error("listTags failed:", e);
    return [];
  }
}
```

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespaceFieldOptions.js
osalama7: Not really, patchedSchema has a different structure, once we patch schema, the original schema gets updated in place.

osalama7:
@cristianoliveira I understand your concerns, This is the last patch before we cut a release. 
We need to focus on Integration and Unit tests next. Everything in this PR is behind a feature flag.

cristianoliveira (approved):
Given the time constraints, let's address these things on a follow up PR

PR #815, opened 2026-02-11
fix(migrations): Improve migraiton fix for user meta

osalama7:
- Fixes for Personal token sql syntax db migration on upgrade with Postgres
```
[GRPC]/service.MigrateService/Migrate cannot update service version for pydio.grpc.token (ERROR: syntax error at or near "not" (SQLSTATE 42601))
```

### DIFF
--- a/idm/oauth/dao/sql/pat-sql.go
+++ b/idm/oauth/dao/sql/pat-sql.go
@@ -52,17 +52,17 @@ func sha256Hash(input string) string {
 }
 
 type PersonalToken struct {
-	UUID              string       `gorm:"column:uuid; primaryKey; type:varchar(36) not null;"`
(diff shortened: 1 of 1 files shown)


idm/meta/grpc/service/service.go
```diff
@@ -63,15 +63,12 @@ func init() {
 					Up:            manager.StorageMigration(),
 				},
 				{
-					TargetVersion: service.FirstRun(),
+					TargetVersion: service.RunAlways().Core(),
```
cdujeu: what's that?

osalama7: An attempt to ensure defaultMetas migration always runs, since the error is ignored in default metas, but perhaps its not necessary.
Will restore FirstRun only

PR #816, opened 2026-02-11
fix(metadata): display form-level errors in UserMetaPanelV2

cristianoliveira:
This commit adds error display for form-level validation failures in the metadata panel.

Detailed Changes:
 - Implementation:
   - Added import for useMetadataContext in UserMetaPanelV2.js.
   - Added context consumption and state access to display form errors.
   - Added error display div that shows when state.errors['form'] exists.
   - Extended save error handling in metadata.tsx to catch and display form-level errors.


cdujeu (approved):
lgtm

PR #817, opened 2026-02-11
fix(api-client): ensure JsonValue is always present in MetaClient

cristianoliveira:
This commit changes the MetaClient to ensure that the JsonValue field is always populated, even when the value is empty.

Detailed Changes:
 - Implementation:
   - Changed MetaClient.js to set JsonValue to an empty string JSON representation when the value is null or undefined.

Relates to #816 for gracefully handling errors on the UI


cdujeu (approved):
lgtm

PR #823, opened 2026-02-17
fix(select): WPB-23512 allow deselecting values in optional select

cristianoliveira:
Now the selector has a little icon for deselect:

Screenshot 2026-02-17 at 13 48 29


frontend/assets/meta.user/res/js/MetaClient.js
```diff
@@ -57,7 +57,8 @@ class MetaClient{
                         meta.Namespace = cName;
 
                         // NOTE: This ensure JsonValue is always present
-                        meta.JsonValue = JSON.stringify(values.get(cName) || '');
+                        const value = values.get(cName);
+                        meta.JsonValue = JSON.stringify(value !== undefined ? value : null);
```
osalama7: Is this needed? stringifying null may return undesired values

osalama7 (approved):

frontend/assets/meta.user/res/js/MetaClient.js
cristianoliveira: Good point, I was experimenting with the empty values, maybe this isn't the way

PR #825, opened 2026-02-18
feat(meta): Patching MetaStore with Entity values

osalama7:
- Add a method in ev dao to fetch entity values for meta uuids
- Extend Search method to fetch entity values if present
- Add unit tests
- Fix Json schema test for missing args

### DIFF
--- a/idm/meta/dao.go
+++ b/idm/meta/dao.go
@@ -85,4 +85,5 @@ type EntityValueDAO interface {
 	LinkMetaValue(ctx context.Context, metaUuid string, valueUuid string) (bool, error)
 	UnlinkMetaValue(ctx context.Context, metaUuid string, valueUuid string) (bool, error)
 	GetMetaEntityValues(ctx context.Context, metaUuid string) ([]*idm.EntityValue, error)
+	GetMetaEntityValuesForMetas(ctx context.Context, metaUuids []string) (map[string][]*idm.EntityValue, error)
 }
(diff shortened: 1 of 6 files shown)


idm/meta/dao/sql/ev-sql.go
```diff
@@ -319,3 +319,47 @@ func (s *evSqlImpl) DeleteEntity(ctx context.Context, entityID string) (*idm.Del
 		RowsDeleted: tx.RowsAffected,
 	}, nil
 }
+
```
ghecquet, line 318: Can you add comments for the linter please ?

idm/meta/dao.go
```diff
@@ -85,4 +85,5 @@ type EntityValueDAO interface {
 	LinkMetaValue(ctx context.Context, metaUuid string, valueUuid string) (bool, error)
 	UnlinkMetaValue(ctx context.Context, metaUuid string, valueUuid string) (bool, error)
 	GetMetaEntityValues(ctx context.Context, metaUuid string) ([]*idm.EntityValue, error)
```
ghecquet: Could we not have one function instead of 2 ?

idm/meta/dao/sql/ev-sql.go
osalama7: Sure thing 👍

idm/meta/dao.go
```diff
@@ -85,4 +85,5 @@ type EntityValueDAO interface {
...
+	GetMetaEntityValuesForMetas(ctx context.Context, metaUuids []string) (map[string][]*idm.EntityValue, error)
```
ghecquet: Could we not have 1 function  instead of 2 for GetMetaEntityValues

osalama7: Their implementation can be simplified, but I might keep the new one to use the return map of entity values by meta-uuid

osalama7: This function GetMetaEntityValues is exposed in the gRPC layer which makes a bit more complicated to refactor, but in the DAO I reused the same query implementation by GetMetaEntityValues so it's only a wrapper at this point.

ghecquet (approved):
I think we can still merge those two functions into one - something we can tackle on later

PR #826, opened 2026-02-18
Feat/handle Array of strings for tags cloud display and input

osalama7:
- Display tag_cloud namespace values from Metastore Array of strings
- Allow multi tags input for tag_cloud json schema
- Use array of strings to accept free input tag cloud
-  add unit test
- add mocks needed

### On Upload
Screenshot 2026-02-18 at 14 34 30

### Side Panel
Screenshot 2026-02-18 at 13 05 03

### On files 
Screenshot 2026-02-18 at 14 34 13

### Next in Search bar


cdujeu (commented):
not sure but I see changes in the **js/fields** (not fieldsv2) library
I think we are "supposed" to left that one untouched and only create new types inside v2 (even if duplicating), ideally coming to a point where we can fully get rid of the v1 
Right @cristianoliveira ?

osalama7:
> not sure but I see changes in the **js/fields** (not fieldsv2) library I think we are "supposed" to left that one untouched and only create new types inside v2 (even if duplicating), ideally coming to a point where we can fully get rid of the v1 Right @cristianoliveira ?

We don't yet have a v2 component for this field, but I'll create a new ticket for it to be refined with Mantine chip  component for colored labels.

cristianoliveira:
Correct, that's the idea. The components that will handle the new types are going to go to fieldsv2/

frontend/assets/__mocks__/pydio/util/lang.js
```diff
@@ -0,0 +1,9 @@
+export default {
+	trim: (str, char) => {
+		if (char) {
+			const regex = new RegExp(`^[${char}]+|[${char}]+$`, 'g');
+			return str.replace(regex, '');
+		}
+		return str.trim();
+	}
+};
```
cristianoliveira, line 9: Why we need to mock trim?

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
```diff
@@ -0,0 +1,399 @@
...
+vi.mock('pydio', () => {
+    const mock = createPydioMock();
+    return { default: mock };
+});
+
+vi.mock('pydio/util/lang', () => ({
+    default: {
+        trim: (str) => str.trim()
+    }
+}));
```
cristianoliveira, line 58: Similar here. But if mocking this is required, shall we try to keep it in one place?

frontend/assets/meta.user/res/js/fields/TagsCloud.jsx
```diff
@@ -0,0 +1,309 @@
...
+            .filter((tag) => !!tag);
+        if (knownTags.length > 0) {
+            tagsList = knownTags.map((tag) => this.renderChip(tag));
+        }
+        const ModernStyles = ThemedModernStyles(muiTheme, {
+            searchRadius: mode === 'popover' ? 8 : null,
+        });
+
+        if (editMode) {
+            const Component = search ? AutoComplete : ModernAutoComplete;
```
cristianoliveira, line 237: I'd rather not mix AutoComplete with TagCloud logic and keep it separated

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
```diff
@@ -0,0 +1,399 @@
...
+        const updateValue = vi.fn();
+        render(React.createElement(TagsCloud, { ...defaultProps, updateValue }));
+        const input = screen.getByTestId('tags-input');
+
+        fireEvent.change(input, { target: { value: '' } });
+        fireEvent.keyPress(input, { key: 'Enter', code: 'Enter', charCode: 13 });
+
+        await waitFor(() => {
+            expect(updateValue).not.toHaveBeenCalled();
+        });
```
cristianoliveira: Hmmm, I'm not sure this is doing what you expect it to do, usually we can't await to something async NOT to happen. But I might be wrong, I believe it's possible if we set some timeout

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
```diff
@@ -0,0 +1,399 @@
...
+        const chips = screen.getAllByTestId('chip');
+        expect(chips).toHaveLength(3);
+    });
+
+    it('parses JSON array string', () => {
+        const props = { ...defaultProps, value: '["a","b","c"]' };
+        render(React.createElement(TagsCloud, props));
+        const chips = screen.getAllByTestId('chip');
+        expect(chips).toHaveLength(3);
+    });
```
cristianoliveira: I believe we agreed to avoid "parsing the data" and work with either arrays or strings. In this case, legacy is handled as comma sperated list (string) and the new types Array of Strings

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
```diff
@@ -37,12 +37,46 @@ import { StringItemsInputProps } from "./CommonInputProps";
...
+        } catch (e) {
+            // If parsing fails, continue with comma-split logic below
+        }
+    }
+
+    if (Array.isArray(value)) {
+        return value.filter((tag) => tag && tag.trim());
+    }
+
+    return (value || '').split(',').filter((tag) => tag && tag.trim());
```
cristianoliveira: This looks over complicated to be fair, that's why I believe is better to have legacy vs new types components that handle data in a consistent form

frontend/assets/__mocks__/pydio/util/lang.js
osalama7: tests are failing on pydio import

cdujeu (approved):

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
osalama7: Its quite simple to handle 2 input variants for array and string, I discussed with Charles we need to implement the v2 component for these types.
Which will be refactored next.

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
osalama7: Yes, but since it's becoming a blocker we need a work around to enable using the new namespace types.

frontend/assets/meta.user/res/js/fields/TagsCloud.jsx
osalama7: This will be reverted later once we have a new component instead of relying on the legacy for display.

frontend/assets/meta.user/res/js/Renderer.tsx
```diff
@@ -85,106 +92,175 @@ export default class Renderer{
...
-        return (<span>{value.split(',').map((tag, index) => {
-            let sStyle = {...tagStyle, ...colorsFromString(tag)};
-            if(index === value.split(',').length -1){
-                sStyle = {...sStyle, marginRight: -4}
+
+        // Parse tags - handle both JSON array and comma-separated string
+        let tags = [];
+
+        // First, try to parse if it's a JSON string
+        if (typeof value === 'string' && value.trim().startsWith('[')) {
```
cristianoliveira: Wait we are repeating this logic here again?

cdujeu: can't we simply JSON.parse() and check the resulting object type instead of '[' token look-up ? 
(i don't know about the duplication topic ;-) )

osalama7: I'll move the mapping to utils and revise.

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
```diff
@@ -0,0 +1,399 @@
...
+        render(React.createElement(TagsCloud, { ...defaultProps, updateValue }));
+        const input = screen.getByTestId('tags-input');
+
+        fireEvent.change(input, { target: { value: 'tag1' } });
+        fireEvent.keyPress(input, { key: 'Enter', code: 'Enter', charCode: 13 });
+
+        await waitFor(() => {
+            expect(updateValue).not.toHaveBeenCalled();
+        });
+    });
```
cristianoliveira: I don't see the duplicated tag input here. Also no need for the await waitFor later, there is nothing async in the key press event 

```suggestion
    it('does not add duplicate tags', async () => {
        const updateValue = vi.fn();
        render(React.createElement(TagsCloud, { ...defaultProps, updateValue }));
        const input = screen.getByTestId('tags-input');

        fireEvent.change(input, { target: { value: 'tag1' } });
        fireEvent.keyPress(input, { key: 'Enter', code: 'Enter', charCode: 13 });

        expect(updateValue).not.toHaveBeenCalled();
    });
```

frontend/assets/meta.user/res/js/fields/TagsCloud.spec.jsx
cristianoliveira:
```suggestion
        expect(updateValue).not.toHaveBeenCalled();
```

cristianoliveira (approved):
I left some comments, it would be nice if we address those before merging. Let's also register follow up tasks if needed for the things we need to address once this is merged

osalama7:
@cristianoliveira I Handled the review comments
- Moved mapper array to strings mapper to utils
- Adding tests
- remove awaits from onKey press events

PR #830, opened 2026-02-20
fix(namespace): Select from value fixes

osalama7:
- Use array of strings json schema for select from values
- Fix duplicate keys in Select Type selection board
- Hide Prompting flags for legacy namespaces


osalama7:
I'll will push the prod dist once other branches are merged to avoid conflicts

frontend/assets/access.settings/res/js/AdminWorkspaces/editor/MetaNamespace.js
```diff
@@ -454,67 +454,72 @@ class MetaNamespace extends React.Component {
...
-                        this.setState({ namespace })
-                        if (v === false) {
-                            this.toggleRequired(namespace)
-                        }
-                    }}
-                    {...ModernStyles.toggleFieldV2}
-                />}
                 
-                {USERMETA_PROMPT_FF &&
+                {(create && !namespace.JsonSchema || namespace.JsonSchema) && (
```
cristianoliveira, line 458: At some point we are going to have to address these complex conditional rendering logic. Probably splitting into small components that can be tested independently

cristianoliveira (approved):

PR #832, opened 2026-02-20
feat(entityvalues): Add new Display json column for entity values

osalama7:
- allows persisting color label for select from values
- fix updating select from values
- refactor create entity values to handle updates
- generate proto bufs


cdujeu (approved):
lgtm

PR #839, opened 2026-02-25
fix(acl): ACL crud test fix

osalama7:
- Remove added acls
- Add acls with simpleCrud function
- Fix simpleCrud usage


cdujeu (approved):
lgtm

PR #840, opened 2026-02-25
fix(metadata): change fieldedit logic to avoid wrong disabling fields

cristianoliveira:
refers to https://wearezeta.atlassian.net/browse/WPB-23701

This commit adds comprehensive tests for the FieldEdit component and fixes wrong disabled props on tag_cloud fields. Detailed Changes:
 - Implementation:
   - Removed duplicate disabled prop from AutoCompleteInput, TagsCloudInput, and tag_cloud field in FieldEdit.tsx to prevent disabling
 - Tests:
   - Added FieldEdit.test.tsx with tests for rendering, tag_cloud field functionality, validation, form state management, accessibility, and edge cases.
   - Renamed CssLabelsField.js to CssLabelsField.jsx for consistency with React component naming.


frontend/assets/meta.user/res/js/fields/CssLabelsField.jsx
cristianoliveira, line 1: This is because of the tests. I didn't want to change the config to handle `js` as `jsx` I think it doesn't make sense

osalama7 (approved):
lgtm

PR #843, opened 2026-02-26
fix(namespace): Add default read policy for namespace list for read only users

osalama7:


cdujeu (approved):
i was really wondering why it would have worked before and not anymore... 
But in fact you did previously modify the existing line to add the "/<.+>" part ;-) 
So yes, we must keep both (namespace and namespace/<.+>)

osalama7:
@cdujeu I added the `/<.+>` part when I introduced this endpoint `/a/user-meta/namespace/jsonschema` , but this pattern excludes (breaks) list `/user-meta/namespace` without any trailing strings for read only users clearly.

PR #849, opened 2026-03-02
WPB-20999: adapt and update EULA used when upgrading from Home to ED

bsinou:
… Home to ED, after validation by legal dpt.


cdujeu (changes requested):
Needs JS recompile - plz check with @cristianoliveira or @osalama7

osalama7 (approved):

PR #850, opened 2026-03-02
poc: bundle js assets on build time

cristianoliveira:
This is a small POC to see how hard would be to remove the minified js from git and generate it at build time


frontend/assets/validate-bundles.sh
```diff
@@ -0,0 +1,283 @@
+#!/bin/bash
```
cristianoliveira, line 1: This script here might be a overkill but I wanted to be able to check and fail the build in case the minified js wasn't properly generated

Makefile
```diff
@@ -6,23 +6,34 @@ GITREV?=$(shell git rev-parse HEAD)
 CELLS_VERSION?=${DEV_VERSION}.${TIMESTAMP}
 GOBIN?=go
 
-.PHONY: all clean build main dev darwin arm win docker start ds licenses
+.PHONY: all clean build main dev darwin arm win docker start ds licenses build-frontend-assets validate-frontend-assets
+
+## Frontend Assets Building
+
+build-frontend-assets:
```
cristianoliveira, line 13: Basically we would just need to add this step to run before the build in the Makefile

PR #851, opened 2026-03-02
fix(namespace): Fixing Namespace resource policies

osalama7:
- Add migration for LoggedUsers read policy

### DIFF
--- a/idm/policy/defaults.go
+++ b/idm/policy/defaults.go
@@ -22,6 +22,7 @@ package policy
 
 import (
 	"context"
+	"slices"
 	"strings"
 
 	"github.com/ory/ladon"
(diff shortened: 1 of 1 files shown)


idm/policy/defaults.go
```diff
@@ -743,6 +743,53 @@ func Upgrade4992(ctx context.Context) error {
 
 }
 
+func Upgrade4993(ctx context.Context) error {
```
cdujeu, line 747: you should use a different approach here: if someone has previously modified their policies in that "bucket", you will override them. 
You would just add the new required one ("rest:/user-meta/namespace/<.+>") I guess

cdujeu (changes requested):

idm/policy/defaults.go
osalama7: Should cover both newly adjusted resources `rest:/user-meta/namespace/<.+>` `rest:/user-meta/namespace`

cdujeu (approved):

PR #855, opened 2026-03-04
fix(metadata): remove setFields action and fix form state handling

cristianoliveira:
Refers to https://wearezeta.atlassian.net/browse/WPB-23832

This commit removes the unused setFields action and ensures form state is handled as a Map to prevent mutation issues.

Detailed Changes:
 - Implementation:
   - Removed setFields action from MetadataActions interface and defaultContext.
   - Changed setFormState action to convert input to new Map to prevent mutation.
   - Updated useEffect hooks to create new Map from node.getMetadata().
   - Added explicit type annotations to setSaving and setShouldSave actions.

### Demo

https://github.com/user-attachments/assets/008d61ba-0be8-496e-b838-587dc3eadc79


osalama7 (approved):

frontend/assets/meta.user/res/js/context/metadata.tsx
```diff
@@ -209,12 +197,18 @@ export const MetadataContextProvider = ({
             });
     }, [node]);
 
+    // Form state initialization once receive
+    React.useEffect(() => {
+        if (!state.jsonSchema) return
+
+        validatorRef.current = buildValidator(ajv.compile(state.jsonSchema))
+        actions.setFormState(new Map(node.getMetadata()))
```
osalama7, line 205: `new Map(node.getMetadata()))` Perhaps this could be initialized only once since its used in both actions

frontend/assets/meta.user/res/js/context/metadata.tsx
```diff
@@ -152,32 +150,31 @@ export const MetadataContextProvider = ({
...
-        setFormState: (formState) => {
+        setFormState: (formState : Map<string, unknown>) => {
             let { isValid, errors } = validatorRef.current(
                 // NODE: To validate and show the currect message "field is required"
                 removeEmptyKeys(formState)
             );
             dispatch({ type: 'set_errors', errors })
 
-            dispatch({ type: 'set_form_state', formState })
+            dispatch({ type: 'set_form_state', formState: new Map(formState) })
```
osalama7, line 162: isn't formState already a map?

cristianoliveira: This is to make a copy of the map and avoid the mutation

frontend/assets/meta.user/res/js/context/metadata.tsx
cristianoliveira: we need both cases because they have different rules, one is more about saving than initialization

PR #859, opened 2026-03-05
refactor(metadata): remove requestToggleClose from search components

cristianoliveira:
This commit removes the requestToggleClose prop from search input components and related tests, as it's no longer needed for the search interface.


osalama7 (commented):
What would be great if all the prettifying changes were done in one separate commit and all the logic was in one PR instead of 2 PRs to be merged into the first PR which actually implements the refactors, so that I can review all the refactoring logic in complete set of changes aiming to achieve one feature or change behavior across components.
One wholesome logic change self contained, self explained and can be tested by checking out one PR otherwise the split into 3 PRs with pretty changes all merged into one is difficult to read or test and validate, I think we need to run the test on the main refactor branch after merging each of the child prs.

cristianoliveira:
I removed all the formatting changes and left only the removal of the prop

osalama7 (approved):

PR #863, opened 2026-03-06
fix(metadata): adds edit/display behavior for datetime fields

cristianoliveira:
Apply the same edit/display style for inputs that use popover for selection


frontend/assets/meta.user/res/js/components/FieldEdit.css
```diff
@@ -6,4 +6,10 @@
 
 .editField.toggable.disabled .mantine-Input-wrapper > .mantine-Input-input {
     background-color: var(--md-sys-color-surface-2);
```
cdujeu: we can remove that line actually, 'background' will take care

cdujeu (approved):
lgtm

PR #866, opened 2026-03-10
feat(metadata): validation flow and info panel save button UX

cristianoliveira:
Refers to https://wearezeta.atlassian.net/browse/WPB-23831

This commit fixes the validation flow in the metadata context and adjusts the info panel save button visibility.

Detailed Changes:
   - Changed InfoPanel to show save button based on mode instead of updateData.
   - Changed metadata context to track mode (idle/editing/saving/invalid) instead of isEditing flag.
   - Added setInitialFormState action to set form state without triggering validation.
   - Updated validation to set mode to 'invalid' when errors exist.
   - Fix: Made onFocus and onBlur props optional in FieldEdit with default noop.
   - Fix: Updated ModalMetaPrompt to pass validateOnSchemaLoad prop and destructure isValid from onDataChanged.
   
 ---
 
https://github.com/user-attachments/assets/0ad9acc2-e50f-4538-884b-fcc49e940e3b


cdujeu (approved):
video lgtm ;-) I did not deep dive in the code
I guess you dbl-check any side-effect on the other contexts? (prompt, meta dialog, etc)

PR #869, opened 2026-03-11
feat(metadata): prefill schema defaults on initial load for PTU

cristianoliveira:
refers to https://wearezeta.atlassian.net/browse/WPB-23532

This commit adds a feature to prefill required and optional schema defaults when a schema is first loaded, if the feature flag is enabled.

Detailed Changes:
 - Implementation:
   - Added a new utility function `mergeOptionalSchemaDefaults` that fills missing required and optional fields with explicit defaults from the schema.
   - Updated `MetadataContextProvider` to conditionally merge defaults on the initial schema load when the `prefillDefaultsOnInitialLoad` prop is true.
   - Added a ref to track whether defaults have been prefilled to prevent reapplication.
   - Updated `ModalMetaPrompt` to enable the new prefill defaults feature.
 - Tests:
   - Added integration tests in `metadata.test.tsx` for the prefill defaults feature, including scenarios for enabled/disabled feature flag and behavior when required fields are cleared.

---

https://github.com/user-attachments/assets/ffd26f5a-ffa8-4d8d-af74-04dd65603b79


frontend/assets/meta.user/res/js/context/utils/validators.ts
```diff
@@ -43,19 +44,63 @@ export const formatSpecialCasesForValidation = (formState: Map<string, any>, jso
...
+    /**
+     * When true, AJV applies schema defaults to missing/empty properties.
+     * Defaults: false
+     * Use case: "Prompt on Upload" to auto-fill metadata fields
+     *
+     * AJV's `useDefaults: "empty"` automatically applies defaults from the schema to any
+     * properties with empty values (null, undefined, "", []) during validation.
+     */
+    applyDefaults?: boolean;
+}
```
osalama7, line 65: should the interface be exported, and is it needed to define one boolean param?

frontend/assets/meta.user/res/js/context/utils/validators.ts
```diff
@@ -43,19 +44,63 @@ export const formatSpecialCasesForValidation = (formState: Map<string, any>, jso
...
-    validator: ValidateFunction<any> | null
-): Validator => (formState: Map<string, any>) => {
-    if (!validator) return { isValid: true, errors: {} };
-
-    let isValid = validator(
-        formatSpecialCasesForValidation(
-            formState,
-            validator.schema
+    schema: AnySchema | null,
+    options?: BuildValidatorOptions
```
osalama7, line 87: ` applyDefaults: boolean` can be simplified

frontend/assets/meta.user/res/js/context/utils/validators.ts
cristianoliveira: The idea is to create named arguments, since javascript doesn't have it in its sintaxe that's the way https://dev.to/producthackers/named-arguments-js-2d99

It's much more readable:
```js
# Bad
foobar("my arg", true, false, false)

# Good
foobar("my arg",  { display: true, sorted: false, another: false })
```

frontend/assets/meta.user/res/js/context/utils/validators.ts
cristianoliveira: how?

frontend/assets/meta.user/res/js/context/utils/validators.ts
```diff
@@ -43,19 +44,63 @@ export const formatSpecialCasesForValidation = (formState: Map<string, any>, jso
...
+): Validator => {
+    if (!schema) return (formState: Map<string, any>) => ({ isValid: true, errors: {}, validator: null })
+
+    const validator = newValidator(schema, options ?? {})
```
osalama7, line 91: I think the validation logic can be done in one place to reduce the boilerplate code, where as the `useDefaults` flag needed for ajv is passed as `true`, I don't see the path where `useDefaults` may receive a null, if not true we can use a default value of the function argument as false to keep it safe.

```suggestion
    const ajv = new Ajv({
        allErrors: true,
        useDefaults: options?.applyDefaults ? "empty" : false,
    })
    addFormats(ajv)
    ajv.compile(schema)
```

osalama7: Suggestion for clarification only,  please don't apply it

frontend/assets/meta.user/res/js/context/metadata.tsx
```diff
@@ -55,8 +54,8 @@ interface MetadataContextType {
     actions: MetadataActions;
 }
 
-const ajv = new Ajv({allErrors: true});
-addFormats(ajv)
+const mapToObject = (map: Map<string, any>) =>
```
osalama7: Can use a [native function](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) for this `Object.fromEntries(entries);`

frontend/assets/meta.user/res/js/context/metadata.tsx
```diff
@@ -245,18 +252,29 @@ export const MetadataContextProvider = ({
...
+        validatorRef.current = buildValidator(state.jsonSchema, { applyDefaults: false })
+
+        let initialFormState = new Map(node.getMetadata())
+
+        // NOTE: This applies defaults values to the form state
+        if (prefillDefaultsOnInitialLoad && state.jsonSchema) {
+            const data = mapToObject(initialFormState)
+            const validator = newValidator(state.jsonSchema, { applyDefaults: true })
+            validator(data)
+            Object.entries(data).forEach(([k, v]) => initialFormState.set(k, v))
```
osalama7: Also not sure about the mapping here, first it was a map, then transformed to an object that is looped over its entries to set `initialFormState`?

frontend/assets/meta.user/res/js/context/utils/validators.test.tsx
```diff
@@ -396,12 +389,12 @@ describe('buildValidator', () => {
...
     // The date-time value should have been converted to ISO string internally
-    // We can't directly observe, but we can trust that Ajv validation passes
+    // We can trust that Ajv validation passes
   })
 
-  it('calls mapErrors with validator.errors', () => {
+  it('calls mapErrors with validation errors', () => {
     // We'll test by checking that errors are mapped correctly
-    const validate = buildValidator(validator)
+    const validate = buildValidator(testSchema, { applyDefaults: false })
```
cristianoliveira: Create a const for `applyDefaults` and use in these tests

frontend/assets/meta.user/res/js/context/utils/validators.ts
```diff
@@ -43,19 +44,63 @@ export const formatSpecialCasesForValidation = (formState: Map<string, any>, jso
...
+                formState,
+                schema
+            )
         )
-    )
-    const errors = mapErrors(validator.errors)
-    return { isValid, errors }
+        const errors = mapErrors(validator.errors)
+
+        return { isValid, errors, validator }
```
cristianoliveira:
```suggestion
        return { isValid, errors }
```

osalama7 (approved):

PR #870, opened 2026-03-12
feat(i18n): add translations for search and admin metadata 

cristianoliveira:
Refers to https://wearezeta.atlassian.net/browse/WPB-23505

This commit adds new i18n keys for search modifiers in the user metadata interface and for boolean values in the admin metadata board.

Detailed Changes:
 - Implementation:
   - Changed MetadataBoard.js to use i18n keys for 'Yes'/'No' display in admin metadata table columns.
   - Changed SearchModifiers.tsx to use i18n keys for search modifier labels.
   - Changed ModalMetaPrompt.js to use i18n key for dialog title.


frontend/assets/meta.user/res/js/TypeEditor.js
```diff
@@ -140,9 +140,17 @@ class TypeEditor extends React.Component {
...
+        const originalM = m;
+        m = (id, fallback = id) => {
+            if (originalM) {
+                const value = originalM(id);
+                if (value && value !== id) {
+                    return value;
+                }
+            }
+            return messageHash['meta.user.' + id] || messageHash['ajxp_admin.metadata.' + id] || fallback;
+        };
```
osalama7: Why was this part changed?

cristianoliveira: I moved it to avoid the cross plugin translation dependency

cristianoliveira: Actually looking further in the code this whole section isn't necessary because the only place that uses this component provides the translation function

osalama7 (approved):

PR #872, opened 2026-03-12
feat(i18n): localize validation error messages

cristianoliveira:
This commit adds localization support for AJV validation error messages. Also makes the messages standard among the fields (see demo below)

 Detailed Changes:
 - Implementation:
   - Added new utility `ajvErrorLocalization.js` that maps AJV error keywords to i18n keys and interpolates parameters.
   - Updated `validators.ts` and `jsonSchema.js` to use the new localization utility.
   - Added new i18n entries for validation messages in `en-us.all.json`.
 - Tests:
   - Updated `validators.test.tsx` to mock the global message hash and verify localized messages.
   - Added comprehensive unit tests for the new localization utility in `ajvErrorLocalization.spec.js`.
   - Updated `jsonSchema.spec.js` to test localization fallbacks.

--- 

https://github.com/user-attachments/assets/f40deb02-612a-45f4-97b3-5167f10f5309


cristianoliveira:
Ok, after some investigation and exploration. There isn't a extension for avj that allows to add custom i18n translations. There is the https://github.com/ajv-validator/ajv-i18n but it a set of pre-defined translations. And there is `ajv-i18n` package as part of Tidelift enterprise subscription. Given that this change has a very small implementation(less than 100 LoC) I believe it should be fine

osalama7 (approved):

PR #874, opened 2026-03-18
fix(s3): add path-style support to fix Ceph RGW S3 compatibility

SJoon99:
## Problem

Ceph RGW (and other self-hosted S3-compatible backends) are effectively
incompatible with Cells when used as a custom S3 endpoint. Every datasource
sync operation and pre-flight connection check fails with `SignatureDoesNotMatch`.

The underlying issue is that Ceph RGW does not support virtual-host-style bucket
addressing (`http://bucket.endpoint/key`). It requires path-style URLs
(`http://endpoint/bucket/key`). When the URL format differs between the request
and the signed string, the signature never matches.

## Root Cause

`path_style` and `force_path_style` are correctly stored in
`DataSource.StorageConfiguration` when a custom S3 endpoint is configured.
However, `FactorizeMinioServers()` only copied `signatureVersion` into
`GatewayConfiguration`, silently dropping these values. They never reached
`MinioConfig.ClientConfig()` and therefore never reached the minio-go
`Options.BucketLookup` field — so the client always fell back to virtual-host
style, breaking Ceph RGW entirely.

## Changes

**`common/proto/object/datasource.go`**
- Add `StorageKeyPathStyle` and `StorageKeyForcePathStyle` constants
- Propagate `path_style`/`force_path_style` through `DataSource.ClientConfig()`
  and `MinioConfig.ClientConfig()`

**`common/config/datasources.go`**
- In all 3 S3 cases of `FactorizeMinioServers()`, propagate `path_style` and
  `force_path_style` into `GatewayConfiguration` alongside `signatureVersion`

**`common/nodes/objects/mc/client.go`**
- In `New()`, set `options.BucketLookup = minio.BucketLookupPath` when
  `path_style` or `force_path_style` is set

**`common/sync/endpoints/cells/transport/config.go`**
- Add `ForcePathStyle` field to S3 transport config struct

**`common/sync/endpoints/cells/transport/mc/s3_client.go`**
- Set `opts.BucketLookup = minio.BucketLookupPath` when `ForcePathStyle` is set,
  applied consistently across all minio client instantiations in the sync transport

**`discovery/install/lib/datasource.go`**
- Set `path_style` and `force_path_style` in `StorageConfiguration` when a
  custom S3 endpoint is configured during installation

**`discovery/install/lib/installer.go`**
- Apply `path_style`/`force_path_style` in the pre-flight S3 connection check
  during install, so the installer validates connectivity correctly against
  Ceph RGW from the start

## Testing

Verified with Ceph RGW as the S3 backend — all datasource sync operations
succeed with no `SignatureDoesNotMatch` errors after this fix.

### DIFF
--- a/common/config/datasources.go
+++ b/common/config/datasources.go
@@ -185,12 +185,18 @@ func FactorizeMinioServers(ctx context.Context, existingConfigs map[string]*obje
 			if !update {
 				newSource.ApiSecret = config.ApiSecret
 			} else {
-				// We may have to update signature version
+				// We may have to update signature version and path-style settings
+				if config.GatewayConfiguration == nil {
+					config.GatewayConfiguration = make(map[string]string)
+				}
 				if sv, o := newSource.StorageConfiguration[object.StorageKeySignatureVersion]; o {
-					if config.GatewayConfiguration == nil {
-						config.GatewayConfiguration = make(map[string]string, 1)
-					}
 					config.GatewayConfiguration[object.StorageKeySignatureVersion] = sv
+				}
+				if ps, o := newSource.StorageConfiguration[object.StorageKeyPathStyle]; o {
+					config.GatewayConfiguration[object.StorageKeyPathStyle] = ps
+				}
+				if fps, o := newSource.StorageConfiguration[object.StorageKeyForcePathStyle]; o {
+					config.GatewayConfiguration[object.StorageKeyForcePathStyle] = fps
 				}
 			}
 		} else if update {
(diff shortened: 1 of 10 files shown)


c12simple:
@SJoon99 

Thank for your work on this topic. I'm having a look at your pull request. Will keep you posted

c12simple:
Hi @SJoon99 
I added a CEPH RGW S3 datasource and didn't encounter any issue.
What I learnt is:

- CEPH RGW S3 by default is path-style
- Cells is using minio-go client. The BucketLookup of this client is by default BucketLookupAuto

_Sets bucket lookup style, whether server accepts DNS or Path lookup. Default is Auto - determined by the SDK. When Auto is specified, DNS lookup is used for Amazon/Google cloud endpoints and Path for all other endpoints._
https://github.com/minio/minio-go/blob/v7.0.65/api.go#L272

Could you please share a concret configuration?

SJoon99:
Hi @c12simple 

You’re right. After further investigation, BucketLookupAuto already handles non-Amazon/Google endpoints (including Kubernetes internal domains) with path-style automatically.

The actual cause of our SignatureDoesNotMatch issue was different:
CELLS_KEYRING defaulted to a local file without a PVC for /var/cells, so the keyring was recreated on every pod restart, breaking secret decryption.

The path_style propagation issue in FactorizeMinioServers() is a separate correctness bug, but not the root cause here.

I’ll close this PR. Thanks for your time.

PR #878, opened 2026-03-20
fix(bookmarks): Fix bookmarking files

osalama7:
- UpdateUserMetaRequestUserMetaOp is exported as IdmUpdateUserMetaRequestUserMetaOp with idm prefix
- Fix imported operation constructor


osalama7:
This will be fixed by restoring the `UpdateUserMetaRequestUserMetaOp` name

PR #885, opened 2026-03-27
chore: WPB-24442 format `meta.user` plugin code

cristianoliveira:
https://wearezeta.atlassian.net/browse/WPB-24442

## Problem

We have introduced prettier to the code, and every time code is save is auto-formatted, which caused some headaches because of merge conflicts. This here is formatting the entire `meta.user` plugging at once, so hopefully the diff in the future will be minimal 

## Approach

Ran command `pnpm run format -- meta.user`


cdujeu (approved):
let's lint! ;-)

PR #895, opened 2026-04-10
fix: add SameSite=Strict to frontend session cookies

cristianoliveira:
Fixes: WPB-23979

https://wearezeta.atlassian.net/browse/WPB-23979


cdujeu:
lgtm but to be tested extensively - could this setting mess up with proxies/ExternalURL settings? 
Beware that add invalid cookie is not necessarily visible when doing some basic tests, as it's involved mostly for the refresh_token operation, so things can look good until you're suddenly logged out after 10mn. 
@c12simple please keep track on that one.

c12simple:
Changing sameSite flag from `None` to `Strict` has no side effect to proxies and refresh_token
- proxies: they are in the middle
- refresh_token: always come from the same site

However, with `strict`, the link won't work. User with already logged tab, in gmail (e.g) clicking on cells link to open the third tab, this tab will request for login. With `lax`, the link may work.

cristianoliveira:
I added some tests to check that too!

cdujeu:
can we add a config for paranoid people? We have a bunch of defaults in the "frontend" section of the configs

cristianoliveira:
A config to enable a security vulnerability? 🤔

cdujeu:
> A config to enable a security vulnerability? 🤔
following Tran's comment, maybe a way to choose between lax and strict? 
if it breaks links it can be really annoying

common/service/frontend/sessions/dao.go
cdujeu, line 1: my 2 cents: create a local sessionOptionsFromConfig(context.Context) function to avoid code duplication? 

This could even go in common/service/frontend/sessions/utils package? to be double-check what it would imply in term of dependencies, but i think that package already pulls common/config...

cristianoliveira: This function, as of now, is only used here in this package, so it doesn't make much sense to move to `utils` once we have the need for broader shared logic, we can move to a shared package. WDYT?

cdujeu: makes sense

cdujeu (approved):

PR #896, opened 2026-04-10
fix: remove unused plugin reload header check from FrontPlugins endpoint

cristianoliveira:
https://wearezeta.atlassian.net/browse/WPB-23975


cdujeu (approved):
lgtm

PR #901, opened 2026-04-13
fix: WPB-23974 Restrict /frontend/enroll access for shared link users 

cristianoliveira:
Refers to https://wearezeta.atlassian.net/browse/WPB-23974

- Split frontend-post policy by profile: standard and shared
- Shared profile users cannot enroll MFA to prevent DoS
- Standard profile behavior unchanged
- Add Upgrade5000 migration for existing databases
- Backward compatible: policy ID and standard profile access unchanged

### DIFF
--- a/idm/policy/defaults.go
+++ b/idm/policy/defaults.go
@@ -854,17 +854,21 @@ func splitFrontendPostPolicies(policies []*idm.Policy) ([]*idm.Policy, bool) {
 // - "frontend-post" for profile:standard (unchanged, backward compatible name)
 // - "frontend-post-shared" for profile:shared (excludes /frontend/enroll)
 func Upgrade4994(ctx context.Context) error {
-	dao, er := manager.Resolve[DAO](ctx)
-	if er != nil {
-		return er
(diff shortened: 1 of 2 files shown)


idm/policy/defaults.go
```diff
@@ -785,6 +886,10 @@ var DefaultsServiceMigrationsAfter4416 = []*service.Migration{
 		TargetVersion: service.ValidVersion("4.9.93"),
 		Up:            Upgrade4993,
 	},
+	{
+		TargetVersion: service.ValidVersion("5.0.0"),
```
cristianoliveira: Not sure what is the rule here, would taht be `4.9.94` instead?

cdujeu: if you want to be able to test it for the next release, make it 4.9.94 indeed (if the current is 93?)

cristianoliveira:
Related to https://github.com/pydio/cells-integration-tests/pull/30

ghecquet (approved):
looks good

osalama7 (approved):

PR #903, opened 2026-04-15
fix: escape dots in MariaDB 10.1 version regex

pedropintosilva:
It looks like the regex intended to match MariaDB 10.1.x versions had
unescaped dots, causing false positives for every MariaDB 10.1X.

    The install wizard incorrectly rejects
    the utf8mb4 charset:

    "This DB version has issues with utf8mb4 charset: Charset is not
    supported for this version of MySQL"

Escape the dots and require digits for the patch component. Use a raw
string literal so the backslashes land literally in the regex.

Tested against MariaDB 10.11.14.

Test covering different name cases added: database_test.go:
cd into cells and then, go test -run TestSpecificVersionsChecks -v
./discovery/install/lib/

```
=== RUN   TestSpecificVersionsChecks

  specificVersionsChecks
    MariaDB 10.11 with utf8mb4 is allowed (regression: unescaped dot false positive) �[32m✔�[0m
    MariaDB 10.1.x with utf8mb4 is rejected �[32m✔�[0m�[32m✔�[0m
    MariaDB 10.1.x with distro suffix and utf8mb4 is rejected �[32m✔�[0m�[32m✔�[0m
    MariaDB 10.1.x with utf8mb3 is allowed �[32m✔�[0m
    MySQL 5.5.x with utf8mb4 is rejected �[32m✔�[0m�[32m✔�[0m
    MySQL 8.0.22 is not supported �[32m✔�[0m�[32m✔�[0m
    non-mysql driver is skipped �[32m✔�[0m

�[31m�[0m�[33m�[0m�[32m
11 total assertions�[0m

--- PASS: TestSpecificVersionsChecks (0.00s)
PASS
ok
github.com/pydio/cells/v5/discovery/install/lib	(cached)
```

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>

### DIFF
--- a/discovery/install/lib/database.go
+++ b/discovery/install/lib/database.go
@@ -191,7 +191,7 @@ func specificVersionsChecks(ctx context.Context, driver string, infos *sql2.Serv
 		return errors.WithMessage(ErrMySQLVersionNotSupported, "version 8.0.22 is not supported")
 	}
 	mysqlMatched, _ := regexp.MatchString("^5.[456].*$", version)
-	mariaMatched, _ := regexp.MatchString("^10.1.*-MariaDB$", version)
+	mariaMatched, _ := regexp.MatchString(`^10\.1\.[0-9]+-MariaDB(-.*)?$`, version)
 	if (mariaMatched || mysqlMatched) && infos.DbCharset == "utf8mb4" {
 		return errors.WithMessage(ErrMySQLCharsetNotSupported, "This DB version has issues with utf8mb4 charset")
 	}
--- a/discovery/install/lib/database_test.go
+++ b/discovery/install/lib/database_test.go
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2018. Abstrium SAS <team (at) pydio.com>
+ * This file is part of Pydio Cells.
+ *
+ * Pydio Cells is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Pydio Cells is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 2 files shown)


pedropintosilva:
without this commit the wizard fails on my setup with MariaDB 10.11.14.

cristianoliveira:
Hello @pedropintosilva 
Thanks for the fix. 
Could you also verify the exact shape of the MariaDB version string returned by the server?

Code reads `SELECT VERSION()` directly in `common/storage/sql/dsn.go`, so the regex should match the real runtime values, not just the simplified semantic version and confirm what MariaDB 10.1 returns on your target setup(s)? 

My concern, the new pattern may be too strict. In many installs the version includes a distro/package suffix, e.g.:
 - 10.1.48-MariaDB - match
 - 10.1.48-MariaDB-0+deb9u2 - match
 - 10.11.14-MariaDB - no match
 - 10.1X-MariaDB - no match

Also, a small test capturing the observed version strings would be nice too.

pedropintosilva:
Thanks @cristianoliveira for the reply, here it is: 

On my target setup (MariaDB 10.11.14 on openSUSE Leap 15.6), `SELECT VERSION()` returns:

```
10.11.14-MariaDB
```

This is the exact string that the old regex `^10.1.*-MariaDB$` was matching as a false positive, the unescaped dots let `.1` and `.*` consume `.11.14`, tripping the 10.1-series utf8mb4 warning on a 10.11 server.

WRT the distro-suffix concern: the new regex `^10\.1\.[0-9]+-MariaDB$` is anchored with `$`, the `10.1.48-MariaDB-0+deb9u2` would NOT match, only exact `10.1.x-MariaDB`. Shall I make it less restrictive: `^10\.1\.[0-9]+-MariaDB(-.*)?$` ?

cristianoliveira:
That makes sense. Your current regex fixes the `10.11.14-MariaDB` false positive. I do believe allowing an optional suffix after `-MariaDB` would be more resilient though. So your suggestion looks good to me, ideally with a couple of regression test cases

pedropintosilva:
Thanks for the review @cristianoliveira , PR updated.

cristianoliveira (approved):

PR #907, opened 2026-04-20
fix: WPB-19708 validate binary content type instead of Content-Type

cristianoliveira:
check for details: https://wearezeta.atlassian.net/browse/WPB-19708

### DIFF
--- a/frontend/rest/binary_validate.go
+++ b/frontend/rest/binary_validate.go
@@ -14,20 +14,25 @@ const (
 
 // allowedBinaryTypes maps detected MIME types to file extensions for binary uploads.
 var allowedBinaryTypes = map[string]string{
-	"image/png":     "png",
-	"image/jpeg":    "jpg",
-	"image/gif":     "gif",
-	"image/webp":    "webp",
-	"image/x-icon":  "ico",
-	"image/svg+xml": "svg",
+	"image/png":    "png",
+	"image/jpeg":   "jpg",
+	"image/gif":    "gif",
+	"image/webp":   "webp",
+	"image/x-icon": "ico",
+}
+
+var binaryContentTypesByExtension = map[string]string{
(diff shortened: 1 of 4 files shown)


frontend/rest/binary_validate.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+)
+
+const (
+	// globalBinaryMaxSize is the max upload size for GLOBAL binaries (5MB, same as avatars).
+	globalBinaryMaxSize = 5 * 1024 * 1024
+)
+
+// allowedBinaryTypes maps detected MIME types to file extensions for binary uploads.
+var allowedBinaryTypes = map[string]string{
```
ghecquet, line 29: That might be a bit of a reductive list. Can you pls check with Tran to see if we have a list of supported content types somewhere ?

frontend/rest/binary_validate.go
```diff
@@ -0,0 +1,82 @@
...
+	n, err := io.ReadAtLeast(content, buf, 1)
+	if err != nil {
+		return "", fmt.Errorf("cannot read file content: %w", err)
+	}
+	buf = buf[:n]
+
+	detected := http.DetectContentType(buf)
+	// DetectContentType returns "type; charset=..." sometimes — take only the MIME part
+	detected = strings.SplitN(detected, ";", 2)[0]
+	detected = strings.TrimSpace(detected)
```
osalama7:
```suggestion
    // http.DetectContentType may append parameters (e.g. "type; charset=...").
    // Use mime.ParseMediaType to extract the clean, lowercased media type per RFC 1521.
    mediatype, _, err := mime.ParseMediaType(detected)
    if err != nil {
    	return "", fmt.Errorf("cannot parse detected content type %q: %w", detected, err)
    }
```

PR #913, opened 2026-04-21
fix: WPB-19707 block unsafe forgot password external links

cristianoliveira:
Refer to https://wearezeta.atlassian.net/browse/WPB-19707


discovery/config/rest/api-root-configs.go
```diff
@@ -90,6 +94,34 @@ func (s *Handler) PutConfig(req *restful.Request, resp *restful.Response) error
 
 }
 
+func validateConfigPayload(fullPath string, parsed map[string]interface{}) error {
```
ghecquet: The idea is valid but it should be done at a lower level in the stack to make sure we cover all different ways of setting configs we can have (command line, grpc). Have a look at this example : 

https://github.com/pydio/cells/blob/9e1a6a04e3cba29b32fdb73d739b53bc5ddd7c46/frontend/rest/plugins-libreoffice.go#L36

common/config/proxies.go
```diff
@@ -57,10 +58,22 @@ func Proxy(store Store) Store {
 	return &proxy{Store: store}
 }
 
+func (p *proxy) Context(ctx context.Context) kv.Values {
```
cristianoliveira, line 61: Found a proxy gap: writes done via `store.Context(ctx).Val(...).Set(...)` bypassed proxy interception. `config.Get(ctx, ...)` uses that path, so the frontend/plugin/core.auth validation proxy was not applied at runtime.

ghecquet (approved):

PR #923, opened 2026-04-24
refactor: WPB-19709 remove imaging dependency from image flows

cristianoliveira:
Refer to https://wearezeta.atlassian.net/browse/WPB-19709 replacing `github.com/disintegration/imaging`


cdujeu (commented):
Okay, that one is pretty touchy ;-) 
I'm more than happy if we can get rid of this lib. That said, I need some more proofs that this is for good ;-) 
Can we produce some comparison between old lib and new implementation:

### Performances
The most important thing - performances comparison between old one and new one for creating a high number of thumbnails, thumbnails from large images, etc.. CPU/ops

### Rendering
Could we eventually compare rendered images with same setting (and same performances). No surprise of a blurry stuff, etc... One important caveats are transparent PNGs - to be checked

### Exif & orientation
Seems to be part of the tests already

cristianoliveira:
After further investigation I found out this PR with a patch https://github.com/disintegration/imaging/pull/175 but more important, a maintained fork https://github.com/boxes-ltd/imaging

cdujeu (approved):

PR #924, opened 2026-04-28
feat/entity-value-policies

osalama7:


idm/meta/rest/handler.go
```diff
@@ -163,8 +164,12 @@ func (s *UserMetaHandler) UpdateUserMetaNamespace(req *restful.Request, rsp *res
 			return errors.WithMessage(errors.InvalidParameters, "user defined meta must start with "+common.MetaNamespaceUserspacePrefix+" prefix")
 		}
 		// Use the helper to validate and get definition
+
 		definition, e := ns.UnmarshallDefinition()
+		fmt.Println(definition)
```
cdujeu, line 169: oops

idm/meta/rest/handler.go
```diff
@@ -163,8 +164,12 @@ func (s *UserMetaHandler) UpdateUserMetaNamespace(req *restful.Request, rsp *res
...
+		// fmt.Println(e)
 		if e != nil {
+			fmt.Println(ns.JsonDefinition)
```
cdujeu, line 172: oopsy

cdujeu (commented):
i think this pr would gain having some detailed description of the why and the how ;-)

PR #928, opened 2026-05-19
Fix(sync): WPB-24026-Partial sync, event and data race bugs

osalama7:
Three critical bugs causing partial file syncs on first run:

1. Unidirectional sync ignores per-root errors
   - runUni() now returns diff.Compute() errors instead of continuing
   - Prevents incomplete patches when a root's diff fails

2. Channel ownership violation in monitorDiff
   - Remove close(done) and close(indexStatus) calls
   - Diff owns channel lifecycle; monitorDiff is consumer-only

3. Concurrent append race in TreeDiff.Compute
   - Protect errs slice with sync.Mutex
   - Two goroutines (left/right loaders) now sync error collection

4. Unit tests ` go test -race ./common/sync/merger -run TestTreeDiffComputeConcurrentErrorsRace -v` ` go test ./common/sync/task --verbose`

### DIFF
--- a/common/broker/fsqueue/adapter.go
+++ b/common/broker/fsqueue/adapter.go
@@ -86,10 +86,10 @@ type fpubQueue struct {
 	basePath string
 	name     string
 
-	closeMu       sync.Mutex
-	closed        bool
-	consumerDone  chan struct{} // Lazily initialized in Consume()
-	closeErr      error
+	closeMu      sync.Mutex
+	closed       bool
+	consumerDone chan struct{} // Lazily initialized in Consume()
+	closeErr     error
 }
 
 // OpenURL implements broker.AsyncQueueOpener.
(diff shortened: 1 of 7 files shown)


common/broker/fsqueue/adapter.go
```diff
@@ -182,35 +202,40 @@ func (f *fpubQueue) PushRaw(ctx context.Context, message broker.Message) error {
 // Consume implements broker.AsyncQueue.Consume.
 // Starts a goroutine that receives messages from the subscription
 // and invokes the callback with decoded broker.Messages.
+
+// Consume implements broker.AsyncQueue.Consume.
 func (f *fpubQueue) Consume(callback func(context.Context, ...broker.Message)) error {
+	f.closeMu.Lock()
+	if f.consumerDone == nil {
+		f.consumerDone = make(chan struct{})
```
ghecquet: can we not initialise on creation ?

osalama7:
Double check if fs queue fails to create a file or creates it with empty content

PR #936, opened 2026-06-03
fix(sync): WPB-26062-Event ordering, metadata handling, and go-routine lifecycle

osalama7:
This PR fixes event ordering and go-routine lifecycle issues in the PubSubEndpoint Watch (subscriber) handler to ensure proper synchronization in diode-based sync systems.

## Event Ordering & Snapshot Authority
- Reorder CreateNode/DeleteNode/MoveNode: snapshot (PathSyncTarget) is now authoritative and updated FIRST before pushing to async queue
- This ensures local state is persisted even if queue push fails
- Warn (but don't fail) if queue Push fails after successful snapshot update; relying on next reconcile cycle to catch up

## Metadata Handling in Watch
- Skip CREATE and UPDATE_PATH events for leaf nodes with incomplete metadata (size==0 && etag=="-1"); wait for UPDATE_META event instead
- This prevents spurious file creations with missing metadata
- Avoids race where metadata arrives after initial event

## Go-routine Lifecycle Management
- Add readWrapper.Close() to wrap reader close with callback execution (mirrors writeWrapper.Close() pattern)
- In Watch cleanup: wait for Consume go-routine to finish via Done() channel with 5s timeout before closing downstream channels
- This prevents "send on closed channel" panics from callback races

## Channel Buffering
- Increase Watch eventChan buffer from 0 to 1000 to absorb event bursts
- Increase errorChan buffer to 100
- Prevents channel sends from blocking during cleanup

## Error Propagation
- Fix GetReaderOn to properly return error (was: `return` → now: `return nil, err`)
- Better context in log messages (e.g., "snapshot updated but queue Push failed")

## Tests
- Add test `pubsub_test.go` 
  - Pub/sub mode node operations (CreateNode, DeleteNode, MoveNode)
  - Error paths (snapshot errors, queue errors)
  - Metadata operations with/without receiver
  - Session management with/without provider
  - Watch functionality (channels, callback, multiple paths)
  - GetWriterOn/GetReaderOn with proper wrapper callback testing
  - 48.6% overall coverage (up from 14.4%) Skipped private functions

Fixes WPB-26062 (pubsub watcher panic and event handling).

### DIFF
--- a/common/sync/endpoints/bus/pubsub.go
+++ b/common/sync/endpoints/bus/pubsub.go
@@ -6,7 +6,10 @@ import (
 	"io"
 	"net/url"
 	"strings"
+	stdsync "sync"
 	"time"
+
+	"go.uber.org/zap"
 
 	"github.com/gobwas/glob"
 	"github.com/pydio/cells/v5/common/broker"
@@ -18,7 +21,6 @@ import (
 	"github.com/pydio/cells/v5/common/sync/endpoints/bus/events"
 	"github.com/pydio/cells/v5/common/sync/model"
 	"github.com/pydio/cells/v5/common/telemetry/log"
-	"go.uber.org/zap"
 )
 
 const (
@@ -276,10 +278,12 @@ func (e *PubSubEndpoint) Watch(ctx context.Context, recursivePath string) (*mode
 		return nil, errors.WithMessage(errors.StatusNotImplemented, "watch not supported in pub mode")
 	}
 
-	eventChan := make(chan model.EventInfo, 1000)
-	errorChan := make(chan error, 100)
+	eventChan := make(chan model.EventInfo)
+	errorChan := make(chan error)
 	doneChan := make(chan bool)
 	wConn := make(chan model.WatchConnectionInfo)
+
+	var cbWg stdsync.WaitGroup
 
 	wo := &model.WatchObject{
 		EventInfoChan:  eventChan,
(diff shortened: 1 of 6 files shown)


common/sync/endpoints/bus/pubsub.go
```diff
@@ -20,6 +18,7 @@ import (
 	"github.com/pydio/cells/v5/common/sync/endpoints/bus/events"
 	"github.com/pydio/cells/v5/common/sync/model"
 	"github.com/pydio/cells/v5/common/telemetry/log"
+	"go.uber.org/zap"
```
ghecquet: I think you don't have the write formatter - should be 

standard

external

organization

common/sync/endpoints/bus/pubsub.go
```diff
@@ -269,8 +276,8 @@ func (e *PubSubEndpoint) Watch(ctx context.Context, recursivePath string) (*mode
 		return nil, errors.WithMessage(errors.StatusNotImplemented, "watch not supported in pub mode")
 	}
 
-	eventChan := make(chan model.EventInfo)
-	errorChan := make(chan error)
+	eventChan := make(chan model.EventInfo, 1000)
+	errorChan := make(chan error, 100)
```
ghecquet: Are we confident about the buffering here ? Are we certain we're draining the chans on cleanup ?

osalama7: I added the buffer to make sure the Consume go routine doesn't panic when it sends on a closed channel after Watch shuts down and tested with `go test ./common/sync/task -v` `TestMonitorDiffMustNotCloseDiffOwnedChannels`, alt. Shall I add a wait group and remove the buffer to close chans properly after callback has returned?

ghecquet (approved):

PR #946, opened 2026-06-24
fix(flows): prevent panic on nil ActionMessage logging

cdujeu:
Prevent logging a nil ActionMessage which leads to panic.


osalama7 (approved):
lgtm

ghecquet:
Reverted as the fix was deviating from the standard behaviou for the MarshalLogObjects. We should either introduce a similar change to all or ensure that such messages are never nil.

PR #964, opened 2026-07-13
fix(admin): improve datasource migration flow and validate DSN connec…

ghecquet:
…tions

  - split datasource migration into clearer source and target selection steps
  - allow choosing or creating the target bucket during migration
  - fix object service gRPC resolution for migration client setup
  - simplify flat/structured migration logic based on datasource config

### DIFF
--- a/cmd/admin-datasource-migrate.go
+++ b/cmd/admin-datasource-migrate.go
@@ -42,11 +42,8 @@ import (
 	"github.com/pydio/cells/v5/common/nodes"
 	"github.com/pydio/cells/v5/common/nodes/models"
 	"github.com/pydio/cells/v5/common/proto/object"
-	registry2 "github.com/pydio/cells/v5/common/proto/registry"
 	"github.com/pydio/cells/v5/common/proto/sync"
 	"github.com/pydio/cells/v5/common/proto/tree"
-	"github.com/pydio/cells/v5/common/registry"
-	"github.com/pydio/cells/v5/common/runtime"
 	"github.com/pydio/cells/v5/common/utils/configx"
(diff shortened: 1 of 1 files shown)


cmd/admin-datasource-migrate.go
```diff
@@ -286,20 +286,50 @@ func migratePickDS(ctx context.Context) (source *object.DataSource, srcFmt, tgtF
...
+		}
+		bucketName = val
+
+		p3 := &promptui.Prompt{Label: "Please enter a location for the new datasource", Default: ""}
+		dsLocation, err := p3.Run()
+		if err != nil {
+			return "", err
+		}
+
+		mc.MakeBucket(ctx, bucketName, dsLocation)
```
osalama7, line 327: is it safe to ignore errors to make a bucket at this stage?

osalama7 (approved):

cmd/admin-datasource-migrate.go
cristianoliveira: I noticed also that `--dry-run` is ignored here and now create a bucket because `migratePickTargetBuckets()` calls `MakeBucket()` unconditionally. Could we pass `migrateDry` into this flow and skip bucket creation during dry-run? It can still return the entered bucket name so the migration preview works, but should not perform `MakeBucket()`.

I believe we need something like
```go
 if bucketName == "Create new bucket" {
       // collect bucketName and location...

       if migrateDry {
           return bucketName, nil
       }

       if err := mc.MakeBucket(ctx, bucketName, dsLocation); err != nil {
           return "", fmt.Errorf("cannot create target bucket %q: %w", bucketName, err)
       }
   }
```

PR #966, opened 2026-07-14
fix(emails): Support footer images in email templates

osalama7:
Add ability to display images in email footers by detecting image URLs in the Copyright field. When a URL to an image is configured, it replaces the default text-based footer with an  tag.

- Upgrade hermes from v2.1.0 to v1.3.0
- Add isImageURL() detector for common image formats
- Implement pydioThemeWithCopyrightImage to render footer images
- Update dependencies (premailer, html2text, goquery)

[Ticket:26212 ](https://wearezeta.atlassian.net/browse/WPB-26212)

Screenshot 2026-07-14 at 10 51 44


broker/mailer/templates/configs.go
```diff
@@ -23,15 +23,26 @@ package templates
...
 	"github.com/pydio/cells/v5/common/config"
 	"github.com/pydio/cells/v5/common/config/routing"
 )
 
+var imageURLPattern = regexp.MustCompile(`(?i)^https?://\S+\.(png|jpe?g|gif|svg|webp)$`)
+
+// isImageURL reports whether s looks like a URL pointing to an image file.
+func isImageURL(s string) bool {
+	return imageURLPattern.MatchString(s)
+}
```
osalama7: Can be replaced with a universal helper

ghecquet: do you need the https in the pattern ? Isn't the extension enough ? we have a utils/std directory where this could go.

osalama7: Can be simplified for sure, I'll follow up on it with another commit.

osalama7: resolved in this commit [05e344bd7e809d8042f0ba801d477c40ced1ed49](https://github.com/pydio/cells/pull/966/changes/05e344bd7e809d8042f0ba801d477c40ced1ed49)

common/utils/std/strings.go
```diff
@@ -1,10 +1,20 @@
...
 	"strings"
 )
 
+var imageExtensions = map[string]bool{
+	".png": true, ".jpg": true, ".jpeg": true,
+	".gif": true, ".svg": true, ".webp": true,
+}
+
+func HasImageExtension(s string) bool {
+	return imageExtensions[strings.ToLower(path.Ext(s))]
```
cristianoliveira, line 19: hmmm

I'm wondering if we need to make this URL-aware

The thing is something like `https://cdn.example/footer.png?v=123` ends up as `.png?v=123` with `path.Ext`, so it will not be recognised as an image

Some concrete cases I can think of:
- `https://cdn.example/footer.png?v=2026-07` for cache busting
- `https://bucket.example/footer.png?X-Amz-Signature=...` for a presigned S3 URL
- `https://cdn.example/footer.svg#logo`

All of these are still image URLs, but would silently render as the normal copyright text

Maybe parse the URL and check `parsed.Path` instead? Would be nice to add tests for these cases too

cristianoliveira (commented):
Left a comment

common/utils/std/strings.go
osalama7: Greg asked to skip url from the original regex, Pushed a [new commit](https://github.com/pydio/cells/pull/966/commits/543847ba334d631a9f138c03ee6a9f1c3910e1c8) to further improve parsing

cristianoliveira (approved):

PR #971, opened 2026-07-18
fix(tag-cloud): WIP - Proposal to fix TagCloud "onlyValuesFromList" behavior + additional flag to force it from the admin (until we attach policies to EValues)

cdujeu:
**DRAFT : not to be merged as is.** 

This PR proposes three things: 

## WPB 27252 - TagsCloudInput onlyValuesFromList not working (Search mode)
 A fix to TagsCloudInput when onlyValuesFromList prop is set (WPB 27252), by simply switching the underlying Mantine component to MultiSelect (searchable) instead of capturing keypress

## TagsCloud default values placeholder
In the admin NS Editor, the Default Values field shown when selecting tagCloud does not display the expected placeholder (the string exists, but wrong prop set, hintText instead of placeholder). I did not create a bug for that one, feel free to do so (and check possible wrong other usage of hintText?).

## TagCloud "closed list" in Edit mode

 A simple implementation to for a TagCloud to use that display mode even in Edit mode. 

1  - Adding a "policyContextEditable" flag in the NSDefinition
This will be used later when we attach policies to EntityValues. We already use that pattern in other API's to evaluate PolicyContextEditable at runtime to inform UX whether the current use is allowed to create values or not. As such, this schema change is not temporary and can be re-used.

2 - Adding a checkbox in the Admin
Until we have policies to evaluate that value properly, simply show a checkbox in the admin NS editor to set it. It  defaults to true (even if not set) to avoid breaking backward compat.

## Still todo
 - Add a new string for the new checkbox
 - Rebase or combine with fix to WPB27253
 - Compile JS
 - The current diff in MetaNamespace.js is not that big, mostly formatting...

### DIFF
--- a/common/proto/idm/meta-ns.go
+++ b/common/proto/idm/meta-ns.go
@@ -58,6 +58,9 @@ type MetaNsDef struct {
 		Items       []ChoiceItem `json:"items,omitempty"`
 		Steps       bool         `json:"steps,omitempty"`
 		EntityItems []string     `json:"entityItems,omitempty"`
+		// Similarly to PolicyContextEditable in other protos, this flag could be computed at runtime  by evaluating
+		// the EntityValues policies, to instruct UX about the current ability to create items or not.
+		PolicyContextEditable bool `json:"policyContextEditable,omitempty"`
 	} `json:"data,omitempty"`
 	GroupName string `json:"groupName,omitempty"`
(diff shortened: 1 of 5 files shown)


cdujeu:
@osalama7 shall I close that draft? i think you've worked on it separately

osalama7:
@cdujeu Yes, implemented in `v5.1-dev`

PR #972, opened 2026-07-20
fix(meta.search): enforce onlyValuesFromList in onChange handler

osalama7:
- Add value filtering in onChangeJoin to respect onlyValuesFromList flag
- Add tests for onlyValuesFromList validation (reject custom, accept list values)

Ticket : [WPB-27252](https://wearezeta.atlassian.net/browse/WPB-27252)

[WPB-27252]: https://wearezeta.atlassian.net/browse/WPB-27252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
```diff
@@ -98,11 +101,11 @@ export const TagsCloudInput: React.FC<TagsCloudInputProps> = ({
...
             splitChars={onlyValuesFromList ? [''] : undefined} // Avoid auto-split on comma
             onKeyPress={(e: React.KeyboardEvent<HTMLInputElement>) => {
-                const { key, target } = e;
-                const { value } = target;
+            const { key } = e;
+            const { value } = e.target as HTMLInputElement;
 
                 if (key === 'Enter' || key === ',') {
-                    if (onlyValuesFromList && !items.includes(value)) return;
+                    if (onlyValuesFromList && items.indexOf(value) === -1) return;
```
cristianoliveira: I think we’re enforcing `onlyValuesFromList` only in some paths here

`onChangeJoin` filters the complete array, but Enter and blur validate only the new value before appending it to `localValue`. If `localValue` already contains something outside the list, we can still commit it together with a valid value

Could we move this filtering into one commit function and use it from `onChange`, Enter, and blur?

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
```diff
@@ -83,8 +83,11 @@ export const TagsCloudInput: React.FC<TagsCloudInputProps> = ({
     }, [name]);
 
     const onChangeJoin = (values: string[]) => {
-        setLocalValue(values.filter((v) => v));
-        onCommitChange(formatTagsArrayToString(values));
+        const filtered = onlyValuesFromList
+            ? values.filter((v) => items.indexOf(v) !== -1)
```
cristianoliveira, line 92: What happens if the user interacts before `dataLoader` finishes?

`items` starts empty, so while the request is pending this filters everything out and commits an empty value. I think we need to either disable the input while loading or wait until the options are ready before validating

It would be good to have a test with a deferred loader for this case

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
```diff
@@ -98,11 +101,11 @@ export const TagsCloudInput: React.FC<TagsCloudInputProps> = ({
             comboboxProps={{ withinPortal: false }}
             splitChars={onlyValuesFromList ? [''] : undefined} // Avoid auto-split on comma
             onKeyPress={(e: React.KeyboardEvent<HTMLInputElement>) => {
-                const { key, target } = e;
-                const { value } = target;
+            const { key } = e;
+            const { value } = e.target as HTMLInputElement;
```
cristianoliveira: Looks like these two lines lost their indentation

cristianoliveira (commented):
Left some comments

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
osalama7: Ran prettier again from `~/assets/`

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
osalama7: Pushed a new commit, preventing input when data loader is not finished would be cleaner.

frontend/assets/meta.user/res/js/fieldsv2/TagsCloudInput.tsx
osalama7: > Could we move this filtering into one commit function and use it from

`onChange` 👍 
`onKeyPress` 
`onBlur`  

The last two can be handled by Maintine, I can demonstrate the custom handling wasn't needed.

cristianoliveira (approved):

PR #976, opened 2026-07-28
fix(WPB-25927): prevent FirstRun migration from overwriting installer configs

osalama7:
Split the migration to:
- Always write structural defaults (secureHeaders, secureCookies)
- Only write plugin param defaults if not already set by the installer

Ticlet: [WPB-25927](https://wearezeta.atlassian.net/browse/WPB-25927)

[WPB-25927]: https://wearezeta.atlassian.net/browse/WPB-25927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/web/service/service_test.go
+++ b/frontend/web/service/service_test.go
@@ -29,40 +29,6 @@ import (
 	"github.com/pydio/cells/v5/common/config"
 )
 
-// firstRunMigrationUp extracts the FirstRun migration Up func from the
-// service definition so tests can invoke it directly without starting a
-// full runtime.
-func firstRunMigrationUp(ctx context.Context) error {
-	// Write structural frontend defaults that are safe to always set on first run.
-	if err := config.Set(ctx, map[string]interface{}{"X-XSS-Protection": "1; mode=block"}, "frontend", "secureHeaders"); err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


frontend/web/service/service_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	. "github.com/smartystreets/goconvey/convey"
+
+	"github.com/pydio/cells/v5/common/config"
+)
+
+// firstRunMigrationUp extracts the FirstRun migration Up func from the
+// service definition so tests can invoke it directly without starting a
+// full runtime.
+func firstRunMigrationUp(ctx context.Context) error {
```
cristianoliveira: The thing is, this is another implementation of the migration rather than the migration we register in `service.go`. It already drifted a bit with `RANDOMSTRING`, which only exists here. So the test can pass while the real FirstRun migration is broken. Can we extract the production `Up` function and use that same function both in the service registration and here?

cristianoliveira (commented):
I think the production change goes in the right direction and matches WPB-25927. I left one concern around the test setup because right now it can stay green even if the real migration changes

frontend/web/service/service_test.go
osalama7: Good point, moved the logic to a helper function.

cristianoliveira (approved):

PR #995, opened 2026-09-16
feat: [WPB-27403] hide collabora action buttons

jose-subero:
This PR introduces a plugin to restrict individual Collabora Online capabilities on Cells admin
the feautres that are now configurable are:

- print
- export
- save
- copy
- access mode
- session termination
-  repair 

Main Ticket: https://wearezeta.atlassian.net/browse/WPB-27403?atlOrigin=eyJpIjoiZTA0MzFmYzMxYjIyNGM0MDlhZTdkNTQ5ZmJlYzVmZWEiLCJwIjoiaiJ9

Feature Tickets:
- https://wearezeta.atlassian.net/browse/WPB-28739?atlOrigin=eyJpIjoiMDU0MjAzODVmZDU0NGVkOWFkODAwMjA0Mjk5OTZlNGYiLCJwIjoiaiJ9
- https://wearezeta.atlassian.net/browse/WPB-28745?atlOrigin=eyJpIjoiMWQwNzJiZWY2Mzg5NDMzNmEwYzM5NjFiNTIxZDFlNTciLCJwIjoiaiJ9
- https://wearezeta.atlassian.net/browse/WPB-28764?atlOrigin=eyJpIjoiMWVkZGY2MDMwNmM2NDRkMWIwNzk1NzRhMmVmOWVjOGYiLCJwIjoiaiJ9


gateway/wopi/fileinfo-builder.go
```diff
@@ -74,12 +76,36 @@ func (dfi *BaseFileInfoResponseBuilder) Build(ctx context.Context, n *tree.Node,
...
 		f.UserCanWrite = false
-	} else {
-		f.UserCanWrite = true
+	case "comment":
+		if f.UserCanWrite {
+			f.UserCanOnlyComment = true
+		}
 	}
 
+	f.EnableOwnerTermination = conf.Val("COLLABORA_DISABLE_OWNER_TERMINATION").Default(false).Bool()
```
cristianoliveira: Maybe I’m missing something, but I don’t think `EnableOwnerTermination` can work with the current owner value. `OwnerId` is always `pydio`, while `UserId` is the current user’s login, so they won’t match for a normal user

Could we get the actual document owner here and add tests for owner and non-owner users? Otherwise I think we should leave this setting out for now

frontend/assets/editor.libreoffice/manifest.xml
```diff
@@ -11,6 +11,13 @@
...
+        <global_param name="COLLABORA_DISABLE_SAVE" type="boolean" label="CONF_MESSAGE[Disable Save / Save As]" description="CONF_MESSAGE[Hide the save button and block Save As for all users in Collabora Online]" default="false" mandatory="false" expose="true"/>
+        <global_param name="COLLABORA_DISABLE_COPY" type="boolean" label="CONF_MESSAGE[Disable Copy]" description="CONF_MESSAGE[Block copying document content to external apps (internal copy/paste still works)]" default="false" mandatory="false" expose="true"/>
+        <global_param name="COLLABORA_DISABLE_MODE" type="select" label="CONF_MESSAGE[Default Access Mode]" description="CONF_MESSAGE[Downgrade the opening mode for all users: edit, comment-only, or read-only]" default="edit" mandatory="false" expose="true" choices="edit|Edit,comment|Comment only,readonly|Read only"/>
+        <global_param name="COLLABORA_DISABLE_OWNER_TERMINATION" type="boolean" label="CONF_MESSAGE[Owner Can Terminate Sessions]" description="CONF_MESSAGE[Allow the document owner to terminate other editing sessions]" default="false" mandatory="false" expose="true"/>
```
cristianoliveira: Small naming thing: `COLLABORA_DISABLE_OWNER_TERMINATION` reads like `true` disables it, but it does the opposite. Could we rename it to `COLLABORA_ENABLE_OWNER_TERMINATION` before this config key becomes something we need to keep?

cristianoliveira (commented):
I found one issue with the owner termination setting

gateway/wopi/fileinfo-builder.go
jose-subero: It seems that you are right,  the OwnerId is hardcoded as "pydio" and never matches the actual UserId, EnableOwnerTermination wouldn't work for any real user. 

So for now i've removed the setting from this PR. 

And we can revisit it in a separate ticket.

frontend/assets/editor.libreoffice/manifest.xml
jose-subero: Agreed on the naming, but since we're dropping the EnableOwnerTermination feature from this PR for now, there is no rename needed for now.

cristianoliveira (approved):

ghecquet (approved):

